PR #11, opened 2020-07-22
Ontology improvements

seanmajorpayne:
- make_network is now split into a main file and class
- reduced redundancy
- added kameron's new updates
- pep8 stylized


knowledge_graph/network_class.py
```diff
@@ -7,32 +7,70 @@ class Network:
...
-        self.result.append((parent.label[0], child.label[0], self.edge_type))
+        self.result.append((parent.label[0], child.label[0], edge_type))
         if child not in self.visited:
             self.visited.add(child)
-            self.node_family.append((child, iter(getattr(child, self.edge_type)))) 
+            self.node_family.append((
+                        child, 
+                        iter(getattr(child, edge_type)),
+                        edge_type
+                        ))
```
rodriguesk, line 41: Replace this section with the following code:

```
            self.node_family.append((
                        child,
                        iter(child.causes_or_promotes),
                        "causes_or_promotes"
                        ))
            self.node_family.append((
                        child,
                        iter(child.is_inhibited_or_prevented_or_blocked_or_slowed_by),
                        "is_inhibited_or_prevented_or_blocked_or_slowed_by"
                        ))
```

knowledge_graph/make_network.py
```diff
@@ -49,18 +47,19 @@ def main(args):
     [give_alias(x) for x in properties]
```
rodriguesk, line 47: **change** the variable `properties` to "obj_properties" in the two lines above here.
AND **add** to this section the following code:
```
annot_properties = list(onto.annotation_properties())
[give_alias(x) for x in annot_properties]
```

PR #20, opened 2020-08-10
Added user scores api

seanmajorpayne:
To run the test, you will need to open two terminal (command prompt) windows. One will run the flask app, the other will run my test file 'test_posting.json'

If this is done correctly, you will get the JSON returned in the console nicely formatted with a 4 space indent.


rodriguesk:
Please merge to master and then delete the branch! Great work Sean and Xavier!

PR #21, opened 2020-08-16
Added dummy scoring code

dmarx:
# Summary

Scoring function API currently assumes some local unimplemented datastore of user information from which user attributes can be retrieved via user_id. Probably a bit over engineered, will likely change to just accepting user features instead.

Assigns random uniform values to all nodes with class "downstream negative effect". 

# Observations

The node class property is currently a hodgepodge of colliding properties. Strongly recommend refactoring to compartmentalize classes in a more semantically meaningful way, e.g:

    # current state
    node['classes'] = ['security societal', 'power resources', 'tradition', 'downstream negative effect', 'test ontology', ...]

   ```
# future state
   node['values'] = ['security societal', 'power resources', 'tradition']
   node['downstream negative effect'] = True # non-optional boolean attributes populated on all nodes, if we so choose
   node.get('test ontology') = True # optional boolean attributes could be implemented like this, i.e. only exist on nodes where they are valued True
```

### CODE REVIEW

File: knowledge_graph/scoring.py [hunk 1]
```diff
@@ -0,0 +1,57 @@
...
+    @property
+    def candidates(self):
+        if not hasattribute(self, '_candidates'):
+            self._candidates = self.get_scoring_candidates(self.g)
+        return self._candidates
+    @staticmethod
+    def get_node_scoring_candidates(g):
+        candidates = []
+        for node_id, node_data in g.nodes(data=True):
+            if 'downstream negative effect' in node_data['classes']:
```
biotom, line 26: Small thing, but should these be extracted as constants instead of being free text in the code?

knowledge_graph/scoring.py
```diff
@@ -0,0 +1,57 @@
...
+    
+    def topk_nodes(self, user_id, top_k=None):
+        candidates = self.get_node_scoring_candidates(self.g)
+        
+        # dummy random scoring
+        random.seed(hash(user_id)) # eliminate stochasticity for now
+        scored = [[c, random.random()] for c in candidates]
+        scored.sort(key = lambda x: x[1], reverse=True)
+        
+        if top_k:
```
biotom, line 44: Should there be a check to see if top_k is an integer?

biotom: And within the bounds of scored? Or have some error handling if top_k is out of bounds?

rodriguesk:
See comments above and repeat PR.

PR #28, opened 2020-08-19
174355066 factorize networkx object

rodriguesk:
- refactorized
- added support for nested classes


seanmajorpayne (approved):
All endpoints from API still run correctly.

These code changes look good. I like that you are using Pep-8 styling 🥇

PR #30, opened 2020-08-19
Docker compose

znurgl:


rodriguesk:
I can't speak to the code, but I tested the docker-compose by following the readme. 
Works well. 
I suggest adding to the readme a little more detail (assume user hasn't used docker before). To get the app to stop I think the user has to CMD+C. Might be worth adding to the readme?

And should document what "Stop" means (like have "Stop the app running from image".

znurgl:
Readme: I just wanted to put something in it, it needs a huge rewrite anyway.

rodriguesk:
Readme is great now!

rolandoldengarm (approved):

PR #31, opened 2020-08-20
Cleanup files

ppython:
1) As discussed we should remove when possible data / generated files from the repo.

>         deleted:    knowledge_graph/Climate_Mind_DiGraph.gpickle
>         deleted:    knowledge_graph/Climate_Mind_DiGraph.json
>         deleted:    knowledge_graph/Climate_Mind_DiGraph.yaml
>         deleted:    knowledge_graph/Climate_Mind_Digraph_Test_Ont.json
>         deleted:    knowledge_graph/output.csv

I let `Climate_Mind_DiGraph.gpickle` as requested but we should come up with another solution.

2) Changed the make_network.py and make_graph.py imports to be relative. Those scripts should be called from the project root directory directly.
Also the generated files will not be in knowledge_graph but in the root also.


dmarx:
When we get to the point where we want to exclude data files from the repo, we should also add those extensions to the .gitignore file to simplify keeping them out after the tracked copies have been deleted (this will block them from being added to commits in the future).

PR #32, opened 2020-08-20
Dash visualization

ppython:
Here's the Dashboard to view our ontology.

You need to have the recent Climate_Mind_DiGraph.gpickle file then,
From the project root, run knowledge_graph/visualize.py

You'll get a flask server running and should be given a URL to check the dashboard.


rodriguesk (changes requested):
For now we will not delete these files. They are needed until we have functioning production code and databases set up.

PR #34, opened 2020-08-25
making the routes more resful <tom>

biotom:
I've changed the route names around to make them more RESTful- I dont have an updated swagger spec yet, but I've raised a ticket with swagger inspector, so hopefully they'll be able to fix whatever issue I'm seeing.

### DIFF
--- a/knowledge_graph/routes.py
+++ b/knowledge_graph/routes.py
@@ -6,29 +6,31 @@ from json import dumps, load
 
 from flask import request, make_response, abort, Response
 
+from typing import Tuple
+
 from knowledge_graph.Mind import Mind
 
-value_id_map = { 
-    1 : "conformity",
-    2 : "tradition",
-    3 : "benevolence",
-    4 : "universalism",
-    5 : "self-direction",
-    6 : "stimulation",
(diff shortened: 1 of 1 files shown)


knowledge_graph/routes.py
```diff
@@ -6,29 +6,31 @@
...
 
 
 @app.route('/', methods=['GET'])
-def home():
-    return "<h1>API for climatemind ontology</h1>"
+def home() -> Tuple[str, int]:
+    return "<h1>API for climatemind ontology</h1>", 200
 
 
 @app.route('/ontology', methods=['GET'])
```
znurgl, line 32: it is usually plural, especially if the response is a list. /ontologies

biotom: Sure, but there's only one ontology- I think 'ontologies' would be misleading.

biotom: My thinking here was that you're querying (via the query string) the ontology

seanmajorpayne:
I think singular for ontology is good.

znurgl (approved):

PR #35, opened 2020-08-28
Feature/174415484 database to store map data

znurgl:


znurgl:
I added the sqlite database file to the repo. I didn't want to do that and because it is in the .gitignore file, it won't be updated in the future. I did this because the ziphazard data is huge and migration would take forever.

biotom:
Probably a dumb question but- will we have problems later if we develop against a SQLite database and try and move to some other SQL for production later?

znurgl:
It is a perfectly valid question. If the prod DB will be a MySQL (for example) then we are going to put it into the docker stack and convert the current sqlite to MySQL.

rodriguesk:
The SQLite queries are the same as SQL. So it won’t affect the code much.
Python is able to connect to and run SQLite so it enables easier to run
locally for now.

znurgl:
I've updated the ZipHazard table. Now it is a cross-reference table with zip_code, hazard_id and value (True, False)

rodriguesk:
I think Andrew already did this. He told me he sent you the wrong table at first and then sent another one? Are we using the new table he provided? There were errors in his first 2 tables too I think.

znurgl:
First two lines of zip_nodes:
``` 
 ,http://webprotege.stanford.edu/RO1J1OifvuO602qTIrSXdB,http://webprotege.stanford.edu/R9vkBr0EApzeMGfa0rJGo9G,http://webprotege.stanford.edu/RJAL6Zu9F3EHB35HCs3cYD,http://webprotege.stanford.edu/RcIHdxpjQwjr8EG8yMhEYV,http://webprotege.stanford.edu/RDudF9SBo28CKqKpRN9poYL,http://webprotege.stanford.edu/RLc1ySxaRs4HWkW4m5w2Me

19903,True,False,,,,
```

andrew-banister:
@znurgl I don't think we need to store **zipHazard** or **hazard** tables in the database. These tables are for preprocessing. We only need to use **zipNodes**. I'm not seeing the code for this, but maybe I'm not familiar enough with github.

znurgl:
@andrew-banister I've exploded that table (as @rodriguesk requested) so now it is a cross-table with zip, hazard_id and value. It is the same as zip_nodes but more flexible.

dao/models.py
```diff
@@ -0,0 +1,14 @@
...
+    zip = db.Column(db.Integer, primary_key=True)
+    hazard = db.Column(db.String, primary_key=True)
+    value = db.Column(db.Boolean)
+
+class Hazard(db.Model):
+    hazard_type = db.Column(db.String, primary_key=True)
+    iri = db.Column(db.String)
+    ref_label = db.Column(db.String)
+    description = db.Column(db.String)
+
```
andrew-banister, line 14: Please remove ZipHazard (**lkp_zip_hazards**) and have just **lkp_zip_nodes**. The backend should just need the IRI's, not the hazard_name. We could keep Hazard (**brg_hazards_nodes**) there for reference since the table is small, but it will not be used by the backend.

db/README.md
```diff
@@ -0,0 +1,11 @@
+# Add Zip Nodes to the database
+
+You can use DB Browser for SQLite to insert zip nodes into a temp database. Let's call it zip_hazard_temp. After a successful
+bulk insert, you can extract the data into zip_hazard table which is a zip code and hazard id table with the corresponding Boolean value. 
+
```
andrew-banister, line 5: I feel like we should have a package that writes the csv from pandas into SQL. I've never worked with DB Browser, but it doesn't seem like a real pipeline where I can change the csv file and run a script to update the table. Kameron and I did this in R using DBI https://db.rstudio.com/dbi/ , there should be something equivalent in Python. Let me know if you need any help.

PR #40, opened 2020-09-05
174682712 process new ontology

rodriguesk:
Made new scripts that will get used in the production pipeline. 

Scripts added process a new OWL file (given its absolute path) and then use it to update the files used downstream by the climate mind backend.

Added a script to call the visualize script from the backend directory.

Updated Readme file.


rodriguesk:
@znurgl Does the Docker file need to be updated? Also, when trying to build I think there's an issue with some other script that uses make_network.py and has an issue with line 5: ```from network_class import Network```. The code I wrote needs it to be like this, but for some reason some other code wants it to look like ```from knowledge_graph.network_class import Network```? Can we change this other code so it's unified?

README.md
```diff
@@ -46,3 +46,36 @@ Start the container
...
+3. New files will be generated and appear in the climatemind-backend directory. These files are also listed in the git.ignore file, so don't worry about them getting pushed accidentally to the git repo.
+
+
+## Visualizing the ontology
+
+Follow these instructions to visualize an OWL file using the Dash dashboard generated by the visualize.py script.
+
+1. Download the new OWL ontology file (if haven't already) and make note of the absolute path. DO NOT SAVE the OWL file in the Climate Mind directory in order to avoid accidentally pushing it to the repo.
+
+2. From the climatemind-backend directory run the process_new_ontology_and_visualize.py with the absolute path to the new OWL file from step 1.
```
rodriguesk, line 69: Note how to  cd to the directory so others can use who aren't familiar with bash.

rodriguesk (commented):
Should I combine the process_new_ontology_and_visualize.py script with process_new_ontology_file.py and just add a binary argument flag for running the visualization part after processing the new ontology files?

andrew-banister (approved):

PR #41, opened 2020-09-06
removed username from the scores endpoint path <tom>

biotom:


seanmajorpayne (approved):
Looks good Tom. Thank you!

PR #43, opened 2020-09-07
Sqlalchemy database

seanmajorpayne:
I've added the sqlalchemy database for users and scores. This is now under the models file.

I've also updated a few other things that were causing issues (such as circular imports)

Namely, the Config is now under the main directory, rather than nested in knowledge_graph

Flask shell has been updated for easy access to the database. You can now run "flask shell" and make entries and deletions through the DB that way.

Included in the User model are a couple of password hash functions which should be useful when we need to start storing passwords.


znurgl (approved):

rodriguesk:
@seanmajorpayne Please see https://github.com/ClimateMind/climatemind-backend#processing-any-new-climate-mind-owl-ontology-file

The files Climate_Mind_Digraph_Test_onto.json and Climate_Mind_DiGraph.gpickle should not go in knowledge_graph. The files should go in the main repo where the climate_mind_ontology is. Also note that they are included in the git.ignore file so they shouldn't be updated in the repo.

seanmajorpayne:
@rodriguesk I believe I've successfully removed them. I will merge into master now.

PR #37, opened 2020-09-02
adding finalised draft

Expan75:
Hey ya guys!

Here's the final draft. It does not include beautification standards except for PEP8 but I guess we can just add it in later :)

PR #49, opened 2020-09-12
Add black formatter

mncharlton:
The style of code was beginning to vary between developers, this PR introduces a formatter ([Black](https://pypi.org/project/black/)) to make things nice and consistent.

Black will run during CI and if it's formatting isn't met then CI will fail (I've tested this by pushing a commit that caused it to fail, then reverting it)

You can run Black locally to see which files need formatting using `python3 -m black --check ./`
You can get Black to automatically format your files using `python3 -m black ./`

The only real changes I made in this PR are to the circleCI config and to add Black to requirements.txt, plus the readme. All the other changes come from this commit - https://github.com/ClimateMind/climatemind-backend/pull/49/commits/222c3d05484d1b5eeab36bbc44df270750be53d3 - and are caused by the autoformatting being applied across the other files.

@rolandoldengarm - specifically on the CI side, I've not set up CircleCI for a python project before - does what I've done look ok to you?


rodriguesk (approved):
Looks good! I think changes to the contributor guide should also be made. But those changes would have to be added to the contributor guide branch because the guide hasn't been merged to master yet: https://github.com/ClimateMind/climatemind-backend/pull/37

rodriguesk:
@mncharlton feel free to merge at your earliest convenience :)

mncharlton:
^ the new `pygraphviz` package wouldn't install in circleci, so I've changed it so this job just installs black, it's actually better as it'll be quicker and we're not dependant on anything else for linting.

PR #52, opened 2020-09-17
DRAFT, DO NOT MERGE- Using db for routes

biotom:
@seanmajorpayne does it look like I'm on the right track with this? Apologise- merging into master introduced a bunch of new files, but the relevant one is 'persist_scores.py'. I'm working on testing it at the moment.


seanmajorpayne:
@biotom Yes, this looks good to me. That is the correct way to use sqlalchemy classes. Nice work!

seanmajorpayne:
@biotom Can you review my changes to the database structure and feel free to merge this if it looks good! Thank you for your insights and brainstorming 👍

rodriguesk (changes requested):
Please remove the extra files and test this using docker.

Can do so in docker by 

docker-compose up -d

CLIMATEMIND_ID=$(docker ps -lq)

docker exec -it $CLIMATEMIND_ID python Python_script_name_goes_here.py python_script_arguments_go_here...

Can also do so using ‘docker-compose up’ for testing API things

PR #53, opened 2020-09-19
Lrf processor

seanmajorpayne:
@rodriguesk Added script to process LRF CSV and save into database. These should work even if duplicate information is uploaded or new CSVs are added later in the same format.

One thing I haven't considered as of right now is how to deal with a situation where an LRF used to be True, but later changes to False. I can always update this script when that becomes a potential issue.

Database structure has been updated as many-to-many relationship to reduce redundancy.


rodriguesk:
From how I understand it, each new csv is expected to completely overwrite the previous version. That’s how Andrew is envisioning it and me too. Because it has to be in sync with the ontology version. @andrew has already made new updates and changed the data to be associated with new Iri and better processed the map data to be more accurate (which changed a lot of values).

seanmajorpayne:
@rodriguesk Okay, I will make a couple quick adjustments. I think that should be an easy fix.

seanmajorpayne:
@rodriguesk Okay, I've updated this. Now everything from the database is deleted at the start of the process.

rodriguesk (changes requested):
What are all the extra files for? I think these might have been from @biotom?  And I think there’s some Jupyter notebook things too. Any requirements should be added to requirements.txt and scripts can be tested in the docker container using the following:

docker-compose build

docker-compose up -d

CLIMATEMIND_ID=$(docker ps -lq)

docker exec -it $CLIMATEMIND_ID python Python_Script_Name_Here.py script_args_go_here...

seanmajorpayne:
Extra files are from the database. There are a few files that are included
from Tom's pull request as well.

seanmajorpayne:
@rodriguesk I've removed the venv, which has removed all of those extra files. It's also built and running in Docker. Should be good to go.

rodriguesk (approved):
Great work!

PR #56, opened 2020-09-21
174654774/add api tests to ci

mncharlton:
Adds a new step to the CI pipeline to run the API tests on each build - idea being that if the tests fail then you've changed something that the frontend is relying on.
Currently the only tested endpoint is /questions - that will change in the future.


rodriguesk:
What's the status on this? I'd like to not have PRs that are still in active development. It gets really confusing for me.

rolandoldengarm (approved):
Great work @mncharlton !

PR #57, opened 2020-09-22
Added personal-values route

seanmajorpayne:
You can access this route at:

`/personal_values?session-id=1`

Returns top three values and their short descriptions. Since I couldn't find a full list of short descriptions anywhere, I've created temporary ones.

### DIFF
--- a/knowledge_graph/routes.py
+++ b/knowledge_graph/routes.py
@@ -4,9 +4,12 @@ from typing import Tuple
 
 from flask import request, make_response, Response, send_from_directory
 
-from knowledge_graph import app
+from knowledge_graph import app, db
 from knowledge_graph.persist_scores import persist_scores
 from knowledge_graph.score_nodes import get_user_nodes
+from knowledge_graph.models import Scores
+
+from collections import Counter
 
 import uuid
 
@@ -21,6 +24,19 @@ value_id_map = {
     8: "achievement",
     9: "power",
     10: "security"
+}
+
+score_description = {
+    "conformity": "Compliance with rules, laws and formal obligations (Avoidance of violating formal social expectations)",
+    "tradition": "Maintaining and preserving cultural, family and/or religious traditions",
+    "universalism": "You encompass appreciation, tolerance, and general acceptance of the nature of things around you.",
+    "benevolence": "Promoting the welfare of one’s in-groups by being trustworthy and reliable",
+    "self-direction": "Freedom to determine one’s own actions",
+    "stimulation": "Excitement, novelty, and change",
+    "hedonism": "Pleasure or sensuous gratification",
+    "achievement": "Success according to social standards",
+    "power": "Control over people and resources",
+    "security": "Safety, stability and order (security) in the wider society"    
 }
 
 # Swagger Stuff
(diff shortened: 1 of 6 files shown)


biotom:
Looks good! Do you have any tests of the view, or API level tests for this endpoint?

knowledge_graph/routes.py
```diff
@@ -152,6 +168,29 @@ def receive_user_scores() -> Tuple[Response, int]:
...
+
+@app.route('/personal_values', methods=['GET'])
+def get_personal_values():
+    """ Given a session-id, this returns the top three personal values for a user
+    
+    """
+    session_id = int(request.args.get('session-id'))
+    
+    try:
+        scores = db.session.query(Scores).filter_by(session_id=1).first()
```
biotom, line 181: Why are we filtering by session_id = 1 here? shouldn't it be by the session id provided?

knowledge_graph/routes.py
```diff
@@ -152,6 +168,29 @@ def receive_user_scores() -> Tuple[Response, int]:
 
     response = Response(dumps(value_scores))
     return response, 200
+    
+
+@app.route('/personal_values', methods=['GET'])
```
biotom, line 173: Is it more readable to have the session-id parameter as a query string here?

knowledge_graph/routes.py
```diff
@@ -152,6 +168,29 @@ def receive_user_scores() -> Tuple[Response, int]:
...
+def get_personal_values():
+    """ Given a session-id, this returns the top three personal values for a user
+    
+    """
+    session_id = int(request.args.get('session-id'))
+    
+    try:
+        scores = db.session.query(Scores).filter_by(session_id=1).first()
+    except:
```
biotom, line 182: Is it worth specifying some exception types here? For instance- we could return invalid session ID for one exception, but internal server error for others? That might make diagnostics a bit easier in the future

knowledge_graph/routes.py
seanmajorpayne: Whoops, I forgot to change that back. I'll fix.

knowledge_graph/routes.py
seanmajorpayne: What do you mean by this? Can you type some example code here?

knowledge_graph/routes.py
seanmajorpayne: Sure, I'll add some exceptions

knowledge_graph/routes.py
biotom: Sorry, ignore me! I had the syntax mixed up with something else, I thought you could put some parameter in the route name and have flask read it like 'personal_values?session-id=<parameter>', but this is right

knowledge_graph/routes.py
biotom: I'm not seeing these updated exceptions here- there's just a bare 'except'. Am I just not seeing them for some reason?

rodriguesk (approved):
Should probably have the description of the personal values in a separate JSON file so it can change as needed or get pulled from a database or content management system at a later date. 

Overall This is good Though, assuming it works with docker, so I approve.

NickCallaghan:
> You can access this route at:
> 
> `/personal_values?session-id=1`
> 

@seanmajorpayne How would I test this in postman. Is there a working database and a know good session_id to see the response? I've tried session-id=1 and also a valid uuid and got the same response. 

 If session_id is a uuid should:  `/personal_values?session-id=1` not be an invalid call? 

Should there be a different response for record not found as opposed to invalid request format?

seanmajorpayne:
> Should probably have the description of the personal values in a separate JSON file so it can change as needed or get pulled from a database or content management system at a later date.
> 
> Overall This is good Though, assuming it works with docker, so I approve.

Yes, I can pull that out of the main file. Just a temporary quick solution to get front-end moving.

PR #58, opened 2020-09-22
Story to persist user scores to the scores table on the db

biotom:
When a user submits scores via the scores endpoint, this method persists their scores into the scores table. The primary key is a UUID- I'm casting it to an int, which I think is wrong- the smaller size of int vs a UUID means that they will eventually collide. It should be ok for now, so I pushed the code, but if we use a real, in memory DB that should be able to generate a UUID for each entry and we can use that- or something else.


Dockerfile_integration_tests
```diff
@@ -0,0 +1,21 @@
+FROM python:3.8.5-slim-buster
```
znurgl, line 1: @biotom you can extend this file and overwrite the parameters:
Build the first one with a -t parameter, that gives a tag for that. Use that tag in the testing docker file: FROM tagged_name. And only add those that you want to redefine: CMD and WORKDIR.
https://stackoverflow.com/questions/36617904/extending-local-dockerfile

znurgl (approved):

Dockerfile_integration_tests
biotom: Interesting, thanks!

biotom: Are you happy for me to merge this and extend the dockerfile after?

seanmajorpayne:
@biotom You might want to consider this as a future solution for the session-id. It seems UUID is not really secure.

https://stackoverflow.com/a/6092448/13489999

knowledge_graph/models.py
```diff
@@ -38,6 +38,22 @@ def load_user(id):
 
 
 class Scores(db.Model):
+    session_id = db.Column(
+        db.Float, db.ForeignKey("user.id"), primary_key=True, unique=True
```
seanmajorpayne, line 42: primary_keys are by default unique, so that doesn't need to be specified.

seanmajorpayne, line 42: This also needs to get updated as an Integer rather than a float, correct?

seanmajorpayne (changes requested):

biotom:
> @biotom You might want to consider this as a future solution for the session-id. It seems UUID is not really secure.
> 
> https://stackoverflow.com/a/6092448/13489999

That's interesting! But do we mind that much in this case? We just want something that's not going to collide, but we're not really generating session IDs in a cryptographic sense. We should look at real session managment, though- for instance, rolling back DB commits if a session ends early et cetera.

seanmajorpayne:
> > @biotom You might want to consider this as a future solution for the session-id. It seems UUID is not really secure.
> > https://stackoverflow.com/a/6092448/13489999
> 
> That's interesting! But do we mind that much in this case? We just want something that's not going to collide, but we're not really generating session IDs in a cryptographic sense. We should look at real session managment, though- for instance, rolling back DB commits if a session ends early et cetera.
> 
> Maybe like: https://overiq.com/flask-101/sessions-in-flask/

It probably doesn't matter. That makes sense! I'll read this soon.

knowledge_graph/models.py
biotom: Yep, that makes sense- although maybe a string would be even better? I don't think a UUID can really be represented as an int, and casting it into one increases the chances of a collision, since  int is smaller- it should probably be a string. Would changing it to a string break anything that uses the db?

seanmajorpayne:
I think that's fine. As long as it's unique I can't think of any rules it
would violate

PR #62, opened 2020-09-25
CM-20: self documenting routes

biotom:
Could somebody check if this works at the /spec endpoint? I get 'web_1  | ModuleNotFoundError: No module named 'owlready2'' when I try to build.


rodriguesk:
@biotom I dont see Apispec included in the requirements.txt file

NickCallaghan:
> @biotom I dont see Apispec included in the requirements.txt file

yep. causing the docker build to fail for this branch at the minute

biotom:
Yeah, apologies- I've fixed this, I'll push it when I'm not at work

rodriguesk (changes requested):
Should not be requirements.txt in knowledge_graph

biotom:
@NickCallaghan @rodriguesk this is pretty simple so far- I wanted to check I was on the right track before trying to get the formatted comments working

biotom:
> Should not be requirements.txt in knowledge_graph

sorry, must have accidentally created that. thanks for catching!

rodriguesk (commented):
Is this meant for PR now? If so, Please provide screenshot of this endpoint working as expected when run using docker. Meaning show the functionality of auto documentation of API endpoint. I think showing for a recent API that hasn’t been documented yet would be best.

If this is not ready for PR yet, please close and submit when ready for PR.

rodriguesk (changes requested):
@biotom Sorry for all the extra standards we are setting up. Please also make a new branch and have the branch title start with the pivotal tracker id number that the PR associates with. This is so that pivotal tracker auto associates the gitub code with the correct user story. https://www.pivotaltracker.com/story/show/174134249
So for this code, the branch title should start with 174134249

biotom:
I think this should be a lot nicer to use- the docs aren't quite as comprehensive, but to add a new endpoint you just use the @auto.doc decorator, and it presents a nice output with the docstring as a description 

Screenshot from 2020-10-31 15-59-52

NickCallaghan:
Nice @biotom! I think we at lease have a chance of keeping that up to date. Maybe combing that with a postman collection should give us a decent amount of documentations.

 I've just done a push to remove the swagger stuff. doesn't look like we'll be using it. 

As far as I know we don't use /ontology or /get_actions maybe we should remove these from the project altogether

NickCallaghan (approved):

NickCallaghan:
@biotom Recon we should move the docs to the / route? Nothing there anyway, will be easier for folks to find.

biotom:
> @biotom Recon we should move the docs to the / route? Nothing there anyway, will be easier for folks to find.

Could do- could we redirect / to /documentation? It feels nicer to have it as a resource to me.

PR #66, opened 2020-10-04
Owl nx pipeline refactor

dmarx:
I wanted to add some attributes to the networkx graph of the ontology, which I anticipated to be a simple process. As I investigated the processing code to identify where and how to make my planned change, I found the processing procedure extremely confusing and difficult to follow. Additionally, there were several invocations of `eval()`, which are security concerns. The goal of this refactor was to simplify and clarify.

I regression tested this refactor using the included script `test_graph_unchanged.py`, which I used to compare a gpickle output file generated prior to refactoring to one generated afterwards. With the exception of list-valued attributes whose orders may not be consistent, the graphs are equivalent.

----

I do not consider this refactoring complete at this time: I have refactored most of the process that extracts the edge triplets and am currently working on the code that adds node attributes to the graph. I'm already over 20 commits deep, so I figured it would be a good idea to get this reviewed/shipped even though I'm still making changes.

A code smell I want to eliminate is passing the edges from the first part of the procedure to the second via a CSV and reloading the ontology object: both of these are already in attached to the Network class, I want this class to complete the processing to guarantee that both parts of the process are utilizing the same source data.


knowledge_graph/make_graph.py
```diff
@@ -188,48 +141,7 @@ def remove_edge_properties_from_nodes(G, to_remove):
             for node in list(G.nodes[node]["properties"][prop])
             if node not in list(to_delete)
         ]
-
-
-def save_graph_to_pickle(G, outfile_path):
```
rodriguesk, line 193: These functions are supposed to enable us to save the knowledge graph in other formats. Would be good to move them all to utils if possible, but I'm not sure if they all work with our current format of the knowledge graph.

knowledge_graph/make_graph.py
```diff
@@ -188,48 +141,7 @@ def remove_edge_properties_from_nodes(G, to_remove):
...
-    file_path = os.path.join(outfile_path, "Climate_Mind_DiGraph.json")
-    with open(file_path, "w") as outfile:
-        outfile.write(json_graph.jit_data(G, indent=4))
-
-
-def save_test_ontology_to_json(G, outfile_path):
-    file_path = os.path.join(outfile_path, "Climate_Mind_Digraph_Test_Ont.json")
-    with open(file_path, "w") as outfile:
-        outfile.write(json_graph.jit_data(G, indent=4))
+        # DM: uh... won't `node not in list(to_delete)` always evaluate to false?
```
rodriguesk, line 144: I'm pretty sure it doesn't always evaluate to false. Otherwise I wouldn't have wrote it into the code.

knowledge_graph/make_graph.py
```diff
@@ -255,73 +167,6 @@ def get_test_ontology(G, valid_test_ont, not_test_ont):
         remove_non_test_nodes(G, node_b, valid_test_ont, not_test_ont)
 
 
-def get_valid_test_ont():
```
rodriguesk, line 258: We need some utils function to pull out things that are test ontology content for the designer to have a simplified JSON file she can read through to know what content exists in the knowledge graph that could be used in designs. So if deleting this function, then we need to write another function. Also, this gets used in unit testing I think for CI/CD (another reason it would need to be replaced with better code or kept/moved to utils).

knowledge_graph/make_graph.py
```diff
@@ -255,73 +167,6 @@ def get_test_ontology(G, valid_test_ont, not_test_ont):
...
-        "stimulation",
-        "tradition",
-        "universalism",
-        "universalism concern",
-        "universalism nature",
-        "universalism tolerance",
-    }
-
-
-def get_non_test_ont():
```
rodriguesk, line 290: This is again for unit testing to ensure that CI/CD is working properly.

knowledge_graph/make_graph.py
```diff
@@ -255,73 +167,6 @@ def get_test_ontology(G, valid_test_ont, not_test_ont):
...
-        "emissions trading",
-        "mitigation",
-        "solution to indirect adaptation barrier",
-        "solution to indirect mitigation barrier",
-        "solution uncategorized (to do)",
-    }
-
-
-# Read the JSON file back
-def read_json_file(filename):
```
rodriguesk, line 310: we wanted to keep the function read_json_file so that other versions of the knowledge graph could be loaded in without needing to repeat the whole pipeline. This should be moved to utils and not deleted.

knowledge_graph/make_graph.py
```diff
@@ -345,30 +190,26 @@ def makeGraph(onto_path, edge_path, output_folder_path):
     # print(list(default_world.inconsistent_classes()))
 
     # Read in the triples data
-    df = pd.read_csv(edge_path)
+    ## DMARX - csv via make_network.outputEdges()
```
rodriguesk, line 193: The CSV file was more for us to check that our graph traversal code was working correctly. I think it's ok to remove the CSV file step completely from the pipeline. And to move things to the network class as needed like you suggest.

knowledge_graph/make_graph.py
```diff
@@ -345,30 +190,26 @@ def makeGraph(onto_path, edge_path, output_folder_path):
...
+    #          via node_network.result
+    # ... If we've already processed the ontology through the Network object,
+    # why do we need to reload it here?
+    # Can we move add_ontology_data_to_graph_nodes to network_class?
+    df_edges = pd.read_csv(edge_path)
 
-    G = nx.DiGraph()  # There should not be duplicate edges that go the same direction.
-    # If so, need to throw an error.
+    G = nx.DiGraph()
```
rodriguesk, line 200: we still need a check step that throws an error if duplicate edges exist that go the same direction? Just leave a comment for that as a to do if not doing it now.

knowledge_graph/make_graph.py
dmarx: Can you elaborate? If I'm understanding correctly, it's unclear to me what that has to do with where we locate these functions.

knowledge_graph/make_graph.py
```diff
@@ -345,30 +190,26 @@ def makeGraph(onto_path, edge_path, output_folder_path):
...
     remove_edge_properties_from_nodes(G, to_remove)
 
     save_graph_to_pickle(G, output_folder_path)
-    # save_graph_to_gexf(G, output_folder_path)
-    # save_graph_to_gml(G, output_folder_path)
-    # save_graph_to_graphml(G, output_folder_path)
-    # save_graph_to_yaml(G, output_folder_path)
-    # save_graph_to_json(G, output_folder_path)
 
     valid_test_ont = get_valid_test_ont()
```
rodriguesk, line 210: these lines can't work if you deleted these functions, right?

knowledge_graph/network_class.py
```diff
@@ -1,4 +1,7 @@
+import typing
+import owlready2
 from owlready2 import *
```
rodriguesk, line 3: is owlready2 getting imported twice?

knowledge_graph/make_graph.py
dmarx: Ok, I see where I was confused. I'll clarify this to use different variables names to clarify the "node" scope confusion. I'm thinking something like:

    G.nodes[node]["properties"][prop] = [
        value
        for value in list(G.nodes[node]["properties"][prop])
        if node not in list(to_delete)
    ]

knowledge_graph/network_class.py
```diff
@@ -11,47 +14,31 @@ class Network:
...
-        """Adds labels the ontology object in a way that makes them pythonicly accessible through . invocation method.
-
-        Parameters
-        ----------
-        properties: list of ontology property objects to make pythonicly accessible
-        accessible_names = list of properties now accessible via . invocation method
-        """
-        new_names = [self.give_alias(x) for x in properties]
-        return new_names
+        self.obj_properties = [give_alias(x) for x in obj_props]
```
rodriguesk, line 40: I see the benefit of only having the give_alias function in the ontology_processing_utils... but then that would always have to be loaded in in order for the Network class to work right because the network class doesn't have the give_alias built in, but rather calls upon other code? Just trying to make it easier for others who want to use the Network class to not have to go searching for dependency code to include in their scripts to use the Network class.

knowledge_graph/make_graph.py
dmarx: I didn't delete this, I moved it to utils. Additionally, it gets imported into makge_graph.py at line 10, so it's still importable from make_graph as though it was defined here. But while we're discussing this, I was planning on refactoring this a bit more. We don't need a function to return a collection like this, we could just define a variable `TEST_ONTOLOGY = {...}` and import that. But I actually don't think that's the right way to do it either: a flat list of values like this belongs in a text file. My plan is to move these values to a text file and read the file into `get_valid_test_ont()`. This way any downstream functions (e.g. CI/CD) that call get_valid_test_ont will continue to function as expected, but also we'll have clearer change management of what values we consider to be test ontology (e.g. we can modify this list without making it look like we changed any utils functions). Additionally, it will make those values available more generally, so they aren't locked into the python component of the application.

knowledge_graph/make_graph.py
dmarx: To clarify: these functions weren't deleted, just moved to a utils module which make_graph imports from.

knowledge_graph/make_graph.py
dmarx: It wasn't deleted, just moved. See line 103 of knowledge_graph/ontology_processing_utils.py

knowledge_graph/make_graph.py
dmarx: I don't mind outputting a CSV file at this step. The issue I had was that if there was that passing the edges from one phase of the process to another using this file felt like it was creating an opportunity wherein if the first part of the process failed silently, the second part would load the old CSV file and it would be difficult to debug what happened.

knowledge_graph/network_class.py
```diff
@@ -107,28 +91,29 @@ def dfs_for_classes(self, node):
         if classes:
 
             for ont_class in classes:
-                if ont_class != owl.Thing:
+                # if ont_class != owl.Thing:
+                if isinstance(ont_class, owlready2.entity.ThingClass):
```
rodriguesk, line 95: No super sure about this. This dfs is the most complicated aspect of the project. Would need @andrew-banister to review this aspect.

knowledge_graph/make_graph.py
dmarx: As I mentioned above, they were moved to another file from which they are imported at the top of make_graph.py

knowledge_graph/network_class.py
```diff
@@ -156,25 +141,13 @@ def dfs_labeled_edges(self):
...
+                    self.node_family.append(rec)
 
                 while self.node_family:
-                    parent, children, edge_type = self.node_family[-1]
+                    parent, children, edge_type = self.node_family.pop()
                     self.visited.add(parent)
-
-                    try:
-                        child = next(children)
+                    for child in children:
```
rodriguesk, line 151: again, need @andrew-banister to check this.

knowledge_graph/network_class.py
dmarx: I think `from owlready2 import *` was supposed to be deleted. I want to replace this implicit import with explicit imports. Under-the-hood, it's not actually getting imported twice. The first import would import the module and the second import adds the imported functions to the global namespace. It would be ugly, but we could actually repeat `import owlready2` as many times as we wanted after this without any additional computational overhead. It would just be nonsensical clutter.

knowledge_graph/ontology_processing_utils.py
```diff
@@ -0,0 +1,108 @@
+import networkx as nx
+from networkx.readwrite import json_graph
+import os
+
+
+def get_valid_test_ont():
```
rodriguesk, line 6: would be good for these to not be hard coded?

process_new_ontology_file.py
```diff
@@ -14,7 +14,9 @@ def newest(path):
     Output: newest file in folder or None
     """
     files = os.listdir(path)
-    paths = [os.path.join(path, basename) for basename in files]
+    paths = [
+        os.path.join(path, basename) for basename in files if basename.endswith("owl")
```
rodriguesk, line 18: I don't think they always end in .owl 
sometimes it's just a binary file.

knowledge_graph/network_class.py
dmarx: Either way, it should only live in one place. My concern was that we're exposing ourselves to bugs if `give_alias` is changed in one file where it's defined but not the other, but it's expected to behave the same way in both places. Prior to this refactor, `give_alias` was defined on this class but also in `make_network.py`, from which it was imported into `make_graph`. Seemed to me that the version that was attached to the class was supposed to be the same function, so I made it that way. Considering how it was being distributed across three files, I felt it made more sense to move it to utils to be safe. If it makes more sense to keep it attached to the Network class, we could make it a static method. 

The way I think about classes, it's less about not digging through code than it is logical collections of things. If that aliasing is something that is only meaningful in the context of this process, maybe it makes sense to keep it attached to the graph like this. However, if downstream we might want to take a string and convert it to an alias of the kind we expect to find in the graph, then I'm not sure it makes sense to attach it to Network.

rodriguesk (changes requested):
There's just a few comments I had but Just a few things I suggest changing. I think the functionality is all there though. Good work! 

The biggest thing that should change is removing the csv intermediate file.

knowledge_graph/network_class.py
dmarx: Fair. That's partly why I included the regression test: this bit is complicated, and I wanted a concrete way to demonstrate that it's at least generating the same graph.

knowledge_graph/make_graph.py
rodriguesk: I think instead of 'value' should use 'partnerNode'.

knowledge_graph/make_graph.py
rodriguesk: that is for the saving function. Not for the reading function, right? 
`def read_json_file(filename):` I don't see in ontology_processing_utils.py

knowledge_graph/make_graph.py
rodriguesk: Let's take out the part where it reads in the edges from the CSV file. Have it read in edges from the python object directly from the previous step in the pipeline. It can still save a csv for record keeping. Just don't want it to read directly from it, right?

knowledge_graph/network_class.py
rodriguesk: I trust your judgement here. Just please check your code still works with the Docker file pipeline. I think some other parts of the app code use give_alias, maybe the visualize ontology code or the main app code? I think it's smart to move give_alias to ontology_processing_utils just please check no other errors occur when trying to build the docker image and run the API that the front end uses.

knowledge_graph/network_class.py
rodriguesk: The way the current ontology is, not every edge case is tested. @andrew-banister  mentioned this to me when we wrote this double depth first search aspect of the code.

process_new_ontology_file.py
dmarx: The reason I added this is that after freshly cloning the repo, it wasn't the newest file in the folder. `.gitignore` is. 

1. That file should be called ".gitkeep". Pretty sure this one's on me and I actually told you to use ".gitignore"
2. Rather than specifying the file extension of what to look for, we could also just tell it to ignore `fname.startswith('.')`, which captures the current ".gitignore" case and the likely future ".gitkeep" case.

knowledge_graph/make_graph.py
dmarx: yup, exactly what I had in mind!

knowledge_graph/make_graph.py
dmarx: Ah, my bad. Yeah, I think I removed this because it wasn't being used anywhere else in the codebase, and it's just an alias for `json.load(f)`. I can put it back in utils if you want to keep it, but I think it's clearer to just call `json.load`. It's basically self-documenting code. If you feel invoking `read_json_file` adds clarity let's go for it, but my feeling is that stacking layers of function definitions on top of self-explanatory operations just makes the code more confusing rather than less. I definitely don't want to take away convenience functions you find useful, adding it to utils is no big deal.

rodriguesk: Oh ok. If that’s the case then no need to have it.

process_new_ontology_file.py
rodriguesk: Maybe that’s the reason why the visualization script isn’t working for Kay? She gets that the networkx object has 0 nodes and 0 edges.
Do what you think is best. The ontology file name is usually a random name when downloading from web protege.

rodriguesk:
@dmarx ok, so given all our conversations above, can you summarize the consensus we have for what changes you need to make?
If you could make a nice list then it’ll make my second review easy so I can approve it and get your changes into master.

knowledge_graph/make_graph.py
rodriguesk: Please do this and the other changes we talked about that are still needed and then I can approve your PR

knowledge_graph/network_class.py
dmarx: Been running the processing pipeline from within docker for my tests. No errors reported by flask, but I also wasn't actually checking the site. I'll poke around the site and visualization to be sure. I think I was also pretty diligent making sure that any functions I removed weren't being invoked elsewhere in the codebase (by searching the repo in github).

dmarx:
Immediate action items:

* Rename one of the `node` variables in the list comprehension in `remove_edge_properties_from_nodes`
* Change the code that enforces the 'owl' file extension to ignoring dotfiles.
* @dmarx to include light regression testing of front-end and visualization components in review
* @andrew-banister to put a second set of eyes on the DFS implementation.

Not really part of the refactor, but happy to take care of it while I'm mucking around in the pipeline:

* Add a check to ensure edges are unique wrt src->tgt (actually I think this is already guaranteed by invoking `df.drop_duplicates()`, but an explicit check certainly doesn't hurt anything.
* `git mv PUT_NEW_OWL_FILE_IN_HERE/.gitignore PUT_NEW_OWL_FILE_IN_HERE/.gitkeep`

Gonna take a bit more time, might be better to merge this PR first and I'll continue working on this in a separate PR:

* Move rest of networkx graph building to the `Network` class to eliminate need for the intermediate csv output (i.e. refactor the second half of the pipeline).

knowledge_graph/network_class.py
```diff
@@ -107,28 +91,29 @@ def dfs_for_classes(self, node):
...
-                            visited_classes.add(child2)
-                            self.add_class_to_explore(child2)
-
-                except StopIteration:
-                    self.class_family.pop()
+                parent2, children2, edge_type2 = self.class_family.pop()
+                visited_classes.add(parent2)
+                for child2 in children2:
+                    # if child2 != owl.Thing: # ?
+                    if child2 == owl.Thing:  # fr though, what is this?
```
andrew-banister, line 103: owl.Thing is the highest level class that all classes are under. When it gets to owl.Thing it can't go any further

knowledge_graph/network_class.py
```diff
@@ -64,35 +51,32 @@ def add_child_to_result(self, child, parent, edge_type):
...
-                (class_name, iter(self.ontology.get_parents_of(class_name)), "is_a")
-            )  # the class(es) of the ont_class. This could pull classes that are just Restriction classes, so really should add code here that checks the class is found in self.ontology.classes() before adding it to the class_family.
-        except:
-            pass
+            if hasattr(owl_class_obj, obj_prop):
+                val = getattr(owl_class_obj, obj_prop)
+                rec = (owl_class_obj, iter(val), obj_prop)  # why iter()?
+                self.class_family.append(rec)
+
+        parents = self.ontology.get_parents_of(owl_class_obj)
```
andrew-banister, line 76: We had a try in the original code. I like the if hasattr above. Not sure if an if is necessary here as well.

andrew-banister:
network_class.py looks good. Thanks for the updates @dmarx.

andrew-banister (approved):

knowledge_graph/network_class.py
rodriguesk: Looks good. I was confused because I thought isinstance was an Owlready2 function and instance was referring to ontology instance. But it’s not it’s referring just to standard python object type, which makes sense now.

knowledge_graph/network_class.py
rodriguesk: Andrew wants to know why you did not replace the try with an if here for this line. He thinks it might the try should be replaced with an if for this line too.

PR #69, opened 2020-10-12
Api tests for scores endpoint

mncharlton:


rodriguesk:
Hi @mncharlton . Is this failure CircleCI for a specific reason? Might be good for us to chat soon. We haven't in a long time!

PR #70, opened 2020-10-13
[CM-51] Initial tests for ontology endpoint

fe-mason:
Adds basic tests for /ontology endpoint.

This assumes the /ontology response will be empty as per current behaviour. This will fail when the response is no longer empty and we can update the test at that point.


biotom (approved):

rodriguesk:
This endpoint isn’t used or needed for the app we plan to launch.

PR #75, opened 2020-10-20
Fixing database issues

seanmajorpayne:
You'll need to add your IP to Azure to locally test this. It's now returning session IDs.


seanmajorpayne:
@biotom Can you check this on your machine and see why it's not passing your API tests? It's working fine locally.

rodriguesk (approved):
Not sure if the database login and password should be public but given there’s a firewall, I guess that’s okay?

seanmajorpayne:
I'll get this fixed in the morning, change the password and set it to an ENV variable.

PR #80, opened 2020-11-02
Database local initialization

rolandoldengarm:
Removed db initialization from API entrypoint to docker-compose
Also removed `flask db migrate` from API entrypoint as that should only be run manually after changing data model.

### DIFF
--- a/db-init.sh
+++ b/db-init.sh
@@ -1,5 +1,6 @@
 #wait for the SQL Server to come up
-sleep 60
+echo "sleeping DB Init script"
+sleep 30
 
 echo "running DB Init script"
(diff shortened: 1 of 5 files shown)


db-init.sh
```diff
@@ -0,0 +1,6 @@
+#wait for the SQL Server to come up
+echo "sleeping DB Init script"
+sleep 30
```
rodriguesk, line 3: just want to make sure the sleep only happens locally and not with the Azure database. No delay should be needed with Azure database, right?

rolandoldengarm: @rodriguesk yes, this script is only run when running locally and will create an empty database.

The API container will run `flask db migrate` which will then populate the database.

rodriguesk (approved):

PR #86, opened 2020-11-21
More refactoring3

seanmajorpayne:
Please note all changes:

- Removed old code from simple_scoring
- Temporary "actionHeadline" added for frontend
- Rewrote make_graph compute to account for all situations
- Recommended user Nodes now accounts for None by assigning -inf if None found.
- Refactored score_nodes.py for separation of concerns
- Moved json files into their own folder
- Added test_endpoints.py as a manual test for backend until Gabrielle can create more robust version


Dockerfile
```diff
@@ -14,9 +47,11 @@ RUN pip install cython && \
     pip install --upgrade pip && \
     pip install --no-cache-dir -r requirements.txt
 
-EXPOSE 5000 8050
+EXPOSE 1433 5000 8050
```
rodriguesk, line 50: This should not be changed here. It should be changed in the docker-compose.yml

api_tests/test_endpoints.py
```diff
@@ -0,0 +1,33 @@
+import requests
+import json
+
+
+def run_this():
+    url = "http://127.0.0.1:5000/scores"
```
rodriguesk, line 6: This url made need to to change to something like localhost... or 0.0.0.0... for it to work in circleCI. No need to change now but just a heads up for that potential

climatemind.py
```diff
@@ -1,16 +1,17 @@
...
 def make_shell_context():
     return {
         "db": db,
-        "User": User,
+        "User": Users,
         "Scores": Scores,
-        "LRF": Lrf,
-        "Zip": Zip,
-        "Iri": Iri,
+        "Sessions": Sessions,
```
rodriguesk, line 11: Is there supposed to be a comma after Sessions?

docker-compose.yml
```diff
@@ -9,3 +9,21 @@ services:
       - .:/app
     environment:
       FLASK_ENV: development
+      DATABASE_PARAMS: Driver={ODBC Driver 17 for SQL Server};Server=tcp:db,1433;Database=sqldb-web-prod-001;Uid=sa;Pwd=Cl1mat3m1nd!;Encrypt=no;TrustServerCertificate=no;Connection Timeout=30;
+      FLASK_APP: "climatemind.py"
+    command: ["./wait-for-it", "mysql:1433", "--", "python3", "climatemind.py"]
```
rodriguesk, line 14: The “mysql:1433” part should not say MySQL, it should say something like "vc-db:1433" or “db:1433” I think. Not big deal but can be misleading otherwise. maybe “mssql-db:1433” would work, just need to try it before changing.

entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+echo "sleeping entrypoint"
+sleep 30
```
rodriguesk, line 4: This sleep should only happen when it’s localhost. No sleeps should be needed for production docker since the db is always on. 

This might be ok to merge now, but after merging if we notice the webapp makes user wait for 30 seconds a lot of times, this is why.

We only want sleeps happening in the docker-compose.yml. No sleeps should be needed to be triggered in the Dockerfile because they will then be triggered in the production container.

knowledge_graph/routes.py
```diff
@@ -72,9 +58,14 @@ def query() -> Tuple[Response, int]:
...
 @app.route("/questions", methods=["GET"])
+@auto.doc()
 def get_questions() -> Tuple[Response, int]:
+    """
+    Returns the list of available questions that can be presented to
+    the user.
+    """
     try:
-        with open("schwartz_questions.json") as json_file:
+        with open("json_files/schwartz_questions.json") as json_file:
```
rodriguesk, line 68: For computer file paths they can’t be coded with “/“ because it won’t work on windows file systems (will break for devs doing localhost on windows).

Must replace everywhere in the code any file paths with “/“ (this doesn’t apply for urls though!).
Replace each one using the format:

    path_corrected = os.path.join(directory, child_folder, file_name) 

Don’t forget to also import os at the top of the file too

knowledge_graph/routes.py
```diff
@@ -168,75 +165,104 @@ def receive_user_scores() -> Tuple[Response, int]:
...
+        # del scores["session_id"]
+        # del scores["user_id"]
+        # del scores["scores_id"]
+
+        top_scores = sorted(sorted_scores, key=sorted_scores.get, reverse=True)[:3]
 
-        top_scores = sorted(scores, key=scores.get, reverse=True)[:3]
         try:
-            with open("value_descriptions.json", "r") as f:
+            with open("json_files/value_descriptions.json", "r") as f:
```
rodriguesk, line 210: Fix this path using os.path.join()

rodriguesk (changes requested):
Please address all the single line comments made. 
Also please confirm all the files in the migrations folder are identical to those currently in the master branch (otherwise the Azure database will likely break since it’s currently working fine).

climatemind.py
seanmajorpayne: commas after list are okay if they are on their own line. Very handy trick to remember. Makes it easier to add items/comment out other items in the future.

entrypoint.sh
seanmajorpayne: Should only happen once on launch, but we can remember for the future if we have any issues.

seanmajorpayne:
All changes have been fixed (except sleeps which temporarily will not move so we don't break the app)

Dockerfile
```diff
@@ -14,9 +47,11 @@ RUN pip install cython && \
     pip install --upgrade pip && \
     pip install --no-cache-dir -r requirements.txt
 
-EXPOSE 5000 8050
+#EXPOSE 1433 5000 8050
```
rodriguesk, line 50: Roland's production code (his docker-compose.yml file for production) may expect we have EXPOSE 5000 8050 in the Dockerfile. If this merge breaks the production webapp it's likely because of this. Just keep this in mind since removing all the ports from being exposed in the Dockerfile.

api_tests/test_endpoints.py
```diff
@@ -0,0 +1,37 @@
+import os
+import requests
+import json
+
+
+def run_this():
+    target_directory = os.path.join(os.getcwd(), "../json_files")
```
rodriguesk, line 7: I'm pretty sure this still won't work on windows. Literally CANNOT use "/" in all paths without using special libraries. 
To get to previous directory have to use os functions to do so.  I believe what you want is `os.path.join(os.getcwd(), os.getcwd(os.pardir),"json_files"`

api_tests/test_endpoints.py
```diff
@@ -0,0 +1,37 @@
...
+    url = "http://0.0.0.0:5000/scores"
+
+    with open(
+        os.path.join(target_directory, "/sample_user_response.json")
```
rodriguesk, line 11: I'm pretty sure the "/" should be removed here so it looks like `os.path.join(target_directory, "sample_user_response.json")`

knowledge_graph/routes.py
```diff
@@ -72,9 +59,15 @@ def query() -> Tuple[Response, int]:
...
 @app.route("/questions", methods=["GET"])
+@auto.doc()
 def get_questions() -> Tuple[Response, int]:
+    """
+    Returns the list of available questions that can be presented to
+    the user.
+    """
     try:
-        with open("schwartz_questions.json") as json_file:
+        file = os.path.join(os.getcwd(), "json_files/schwartz_questions.json")
```
rodriguesk, line 69: this needs fixing too for good style. 
Should be:
`file = os.path.join(os.getcwd(), "json_files","schwartz_questions.json")`

knowledge_graph/routes.py
```diff
@@ -168,75 +167,105 @@ def receive_user_scores() -> Tuple[Response, int]:
...
+        # del scores["session_id"]
+        # del scores["user_id"]
+        # del scores["scores_id"]
+
+        top_scores = sorted(sorted_scores, key=sorted_scores.get, reverse=True)[:3]
 
-        top_scores = sorted(scores, key=scores.get, reverse=True)[:3]
         try:
-            with open("value_descriptions.json", "r") as f:
+            file = os.path.join(os.getcwd(), "json_files/value_descriptions.json")
```
rodriguesk, line 212: this should also be:
`file = os.path.join(os.getcwd(), "json_files","value_descriptions.json")`

rodriguesk (approved):

rodriguesk (approved):

PR #91, opened 2020-12-22
CM- 287 General myths api endpt 

rodriguesk:
General myths listed out with some short information about them (myth id, myth, what science says) only as response to the /myths endpoint. 

example use: 
http://localhost:5000/myths?session-id=1234


rodriguesk:
Screen Shot 2020-12-22 at 3 20 32 PM
Example local testing showing endpoint works. 
@NickCallaghan I'd like to keep the session-id as part of the required parameter so that the backend can over time change the code to return the myths in a special order that's personalized to the user. Having the session-id is all that is needed to do that at a later time (so no future frontend changes would be needed). 

If you'd like any of the content formatting changes, just let me know and I'll talk with Elle to get it done. 
If you'd like any of the dictionary names changed in the JSON response, just let me know too!

PR #96, opened 2020-12-26
CM-289 separate ontology module from API

bengineerdavis:
Separation between app and knowledge_graph modules are done for now.

Copying some Jira notes, for reference:
Important notes included in recent commit, but need a quick discussion: 

Moving all ontology-related files into knowledge_graph:
-climate_mind_ontology
-process_new_ontology_and_visualize.py
-process_new_onotology_file.py
-test_graph_unchanged.py

Some exceptions that need further clarification:
-Climate_Mind_DiGraph.gpickle is both in the knowledge_graph and also in
the top level directory. Advice needed on which one we should keep/replace.
-climatemind.py needs a better name, unsure if this should remain in top
level directory.
-lrf_processor.py we could revisit making a separate module for data
cleaning and pre-processing for the knowledge_graph module.

Special Instructions:
We need to revisit testing with this new separation in mind, so we know
if we need to change imports, etc.

I’m looking into the migrations folder and whether or not that should be within our “app”/api directory. I have a feeling this could be it's own PR.


rodriguesk (changes requested):
the changes look fine but the code doesn't run everything without errors.
example below:

Error: While importing "climatemind", an ImportError was raised:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/flask/cli.py", line 240, in locate_app
    __import__(module_name)
  File "/app/climatemind.py", line 1, in <module>
    from knowledge_graph import app, db
  File "/app/knowledge_graph/__init__.py", line 8, in <module>
    from config import DevelopmentConfig
  File "/app/config.py", line 5, in <module>
    from knowledge_graph.Mind import Mind
ModuleNotFoundError: No module named 'knowledge_graph.Mind'

bengineerdavis:
> the changes look fine but the code doesn't run everything without errors.
> example below:
> 
> Error: While importing "climatemind", an ImportError was raised:
> 
> Traceback (most recent call last):
> File "/usr/local/lib/python3.8/site-packages/flask/cli.py", line 240, in locate_app
> **import**(module_name)
> File "/app/climatemind.py", line 1, in
> from knowledge_graph import app, db
> File "/app/knowledge_graph/**init**.py", line 8, in
> from config import DevelopmentConfig
> File "/app/config.py", line 5, in
> from knowledge_graph.Mind import Mind
> ModuleNotFoundError: No module named 'knowledge_graph.Mind'

Which script did you run to produce this traceback?

PR #97, opened 2020-12-26
Revet to last know working commit

NickCallaghan:
The POST /scores end point is broken in production this pr will revert to the last know working commit


rodriguesk:
This won’t fix the issue

PR #107, opened 2020-12-29
CM-279 Effect sources added to feed response

rodriguesk:
Effect sources added to feed response.

* added the function get_causal_sources() to score_nodes.py
* get_causal_sources is called in the function simple_scoring in
score_nodes.py
* added processing of causal effect sources to the networkx object
(added code to the function makeGraph in make_graph.py

Note that the code works but that it may look wrong when looking at the
output, but this is because the function set_edge_properties() doesn’t
behave properly. But fixing set_edge_properties() and making needed
changes to remove_edge_properties_from_nodes() functions is out of the
scope of CM-264 branch/PR


rodriguesk:
I also added some code from @y-himanen zipcode branch because it corrected some errors in the main master branch that only get triggered during localhost.

NickCallaghan:
are you able to make sure the strings that are returned are valid urls? there a load returning  when i tested returning Lorem ipsum ?

rodriguesk:
@NickCallaghan the links aren't valid because of an issue with backend source processing code. It's a whole ticket CM-290. I won't be able to get to it until after the zip code story is completed.
This code is correct though. The issue isn't in the scope of this PR.

seanmajorpayne:
@NickCallaghan I've slightly fixed that issue. If no links exist for effectSources it now returns "No sources available at present"

We'll get a better solution turned around soon.

seanmajorpayne (approved):
Looks good to me. Runs as expected.

PR #111, opened 2020-12-29
adding isPossiblyLocal key-value pair to simple_scores.py

bengineerdavis:
This will allow us to return a boolean value for a card based on it's relevance to the user's zip code.


bengineerdavis:
It's important to remember I haven't converted Python True boolean value into the equivalent JavaScript true value, because it otherwise will not pass Black linting.

rodriguesk (approved):

rodriguesk:
@bengineerdavis greatwork!

rodriguesk:
@bengineerdavis the frontend may want the True value as a string. You should ask them. If so just let me know and I'll use administrator privileges to make the change without needing a PR for it.

PR #115, opened 2020-12-30
CM-332-add myth faulty logic description

rodriguesk:


y-himanen (approved):
Tested with postman. Required rebuilding docker.

PR #116, opened 2021-01-01
Cm 240 task cm 298 db changes for lrf data

y-himanen:


rodriguesk (approved):
Looks good. Could add arguments to the add_lrf_table.py to make the "file" be an input argument so that next time the file name could change and the script wouldn't change, but this is fine for now. When the next csv file is made we can do a PR that upgrades this script to be like that.

rodriguesk:
What's the primary key and foreign keys? Is this datatable connected to the other datatables in the database? Also, if there is a table being added, should it somehow be expressed in models.py ? Maybe the add_lrf_table.py should have it's main functions be in models.py and then have add_lrf_table.py call those main functions from models.py? @seanmajorpayne  what do you think?

y-himanen:
- Re: file as an input argument - the script can of course be changed so that the file is an argument on the command in entrypoint.sh. It's not a massive change - but I personally feel that having the file as a simple variable at the top of the script adds some value in terms of readability. 

- Re: primary keys and foreign keys - neither are added automatically using pandas .to_sql. I think the only potentially valuable functionality missing from this is a primary key would force all the values to be unique, so you couldn't accidentally enter the same postal code twice into the lrf table. But considering how updates happen to the table, I don't think this is necessarily critical. As for foreign keys, this kind of is a standalone table in terms of relational databases. The only data matching in other tables is the postal code, but that's not a required value (and will eventually move location from where it is at the moment to the user table?). I also don't know how the app would react if it had to check the lrf table every time to see if the postal code is included there before saving it into the sessions table (performance issues?), nor am I positive what would happen if the user gave an invalid postal code (would it actually fail to make the record? crash the app?). 

- Re: moving anything to models.py. Interested to hear Sean's opinion. Don't know whether that would create clashes with alembic's auto-detection of changes in models.py.

seanmajorpayne:
I will review this in the morning. I'm not at home until late tonight. I'd
like to look over the pandas issue with foreign keys before we merge.

PR #125, opened 2021-01-03
CM 290 bug fix

rodriguesk:
CM-290 removes edge attributes that aren’t sources and adds missing sources to the edge attributes. 
Includes updated processed ontology and networkx object.


seanmajorpayne (approved):
Code looks good. Tested in Postman and confirmed working.

@rodriguesk What is up with the comment from Dmarx? I notice you added another comment but I'm not sure what you're asking about.

PR #127, opened 2021-01-04
version 0.5

rodriguesk:
* removes duplicate solutions
* nulls return empty list for some API responses instead of string message
* bug fix for ontology processing
* updated networkx object (.gpickle file)


rodriguesk (commented):
Good to merge.

PR #142, opened 2021-01-11
CM-253 updated Mind to use networkx

seanmajorpayne:
Not sure exactly what is needed here, but gave it my best shot. There were 0 code comments included and some of the old make_network function imports don't exist now so I don't know what this originally returned.

Endpoint currently returns a specified node (or series of nodes) and all of its data, which is what this did before if I'm not mistaken.

If it needs updating, lemme know.


rodriguesk:
This should not have any OWL file requirements. Only should read in the networkx object gpickle file. The parts that work with the OWL file should be removed.

rodriguesk (approved):

seanmajorpayne:
> This should not have any OWL file requirements. Only should read in the networkx object gpickle file. The parts that work with the OWL file should be removed.

I've removed all ontology references from this file. A couple of notes

The NetworkX Graph is now loaded into the app config. This happens in two possible places
1) If the pickle file exists, it will be set upon loading the app (via the config file)
2) If the pickle does not exist, it will be set after the new ontology file is processed.

When we want to use the NetworkX object now, all we have to do is
G = app.config['G']

Meaning it is only loaded from the file once in the entire app, but can be accessed quickly from any files.

rodriguesk:
I reviewed all this and it looks great! Good job.

This should be fine to merge and another PR can be raised for replacing all the times gpickle is loaded into the app in other places.

PR #143, opened 2021-01-11
feed data table added, migrations complete

y-himanen:
- checked local and cloud - all tables are present and existing data is fine


migrations_azure/versions/d10c6bfdd9aa_put_lrf_table_back.py
```diff
@@ -0,0 +1,65 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "d10c6bfdd9aa"
+down_revision = "65c5753b57e0"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
rodriguesk, line 20: This part here says to add code as needed here. Do we need to adjust any of this?

rodriguesk (approved):
I approve, but there is a comment I made on this PR for a single line. Please review the comment I made.

y-himanen:
those comments are part of the auto-generated script. the auto-scripting isn't perfect and doesn't cover every single use case, so they know that the devs need to go in and adjust at times (just like i did). nothing to worry about.

rodriguesk:
Yes, it just currently says to drop the lrf table upon upgrade or to remove
the index upon upgrade. Is that correct?

y-himanen:
i think you're referring to the migration to put the lrf data back in? it dropped the empty table before putting in a new table with all the values. if so, that's fine. sorry, commenting is a little unclear through here.

rodriguesk:
@y-himanen when you run the next upgrade on the Azure database the following will run:

def upgrade():
    # ### commands auto generated by Alembic - please adjust! ###
This conversation was marked as resolved by rodriguesk
    op.drop_index("ix_lrf_data_postal_code", table_name="lrf_data")
    op.drop_table("lrf_data")
    # ### end Alembic commands ###


Is this the behavior you want?

PR #144, opened 2021-01-12
Cm 369 separation of concerns

seanmajorpayne:
Rewrote scoring, network x, myths, solutions as classes and separated out into grouped files. I think this will be a welcome change for everyone.

Ran tests with Postman and results look the same to me. Please review the output to make sure nothing has been altered due to the code logic. (I'm pretty confident everything is in order).

@rodriguesk I think you should review this.

@bengineerdavis This will be good for you to see before we meet tomorrow. We'll continue from here with the reorganization.


seanmajorpayne:
Oh as a side note, right now we're still loading in the NetworkX graph multiple times. Since the score_nodes function removes nodes from the graph, it's not as simple as loading it once and keeping it stored as a global variable. Myths and Solutions get removed during that process, so in order to do this correctly (at least in the short term) we need to load the graph separately for each class.

rodriguesk:
We just need to do G.copy() then

rodriguesk (approved):

PR #146, opened 2021-01-12
Cm-371 fix missing myth iris

rodriguesk:


seanmajorpayne (approved):
Tested in Postman and with creating new ontology.

Creating new ontology failed due to an import that I messed up in my own code. This is now fixed and working.

Approved.

PR #148, opened 2021-01-13
Cm 296 create app tests directories

bengineerdavis:
This should give us a good starting place to structure our tests and make testing a regularly habit across our backend team.


bengineerdavis:
This should actually have been merged with CM-252, which I just did, sorry for the confusion on this.

seanmajorpayne:
@bengineerdavis Nice, awesome you're getting a PR up. If a PR fails the lint test, it tells me that the developer hasn't reviewed their own PR before requesting a review.

In order to pass lint tests, you need to run black on the directory. You can see instructions on how to do this in the readme.

Also please make sure to run postman and verify the everything is working before having someone else review.

Lastly, I notice you put a lot of commits for small changes. We have to pay for circleCI credits which are used every commit. If you can push less commits in between PRs that is ideal.

Thanks!

rodriguesk:
As open source we have lots of circle ci credits, but there is a limit. It’s a high limit, but we just all have to be sustainable with commits (commit when you would like to test).

bengineerdavis:
Hi @rodriguesk @seanmajorpayne thank you for the head's up--I will fix this and make sure to be more efficient with using CI.

bengineerdavis:
I was having some trouble from doing a regular eye test what the issue was, and running my local system install of the Black linter also didn't see any problems with that file. I figured there's something specific about the way we customized Black within the docker environment that's then why there's an issue, so I run ```docker exec -it $WEB black add_lrf_table.py``` and that fixed it.

bengineerdavis:
Looking for another pair of eyes to merge this @seanmajorpayne @rodriguesk

PR #149, opened 2021-01-13
Readding kamerons changes

seanmajorpayne:
Temporary fix with some duplicate code. I will update this again tonight when I get home, but want to have it working for the front-end.


rodriguesk (approved):
remove the print statement and resolve the merge conflicts and anything else needed you notice.

No need to import pdb when using breakpoints(). So you can remove the import pdb. Please look over the code another time that changes to ensure it's all necessary.

PR #152, opened 2021-01-13
Cm 329 bug fix subscribe code

y-himanen:


rodriguesk (approved):
Might be good to later rename the table 'subscriptions' but signup is fine for now.

PR #156, opened 2021-01-15
CM 367 return objects for errors (and small readme update)

seanmajorpayne:
Updating the errors so that they return objects as Nick requested

Added PDB information to readme

### DIFF
--- a/knowledge_graph/routes.py
+++ b/knowledge_graph/routes.py
@@ -60,69 +60,21 @@ def home() -> Tuple[str, int]:
     return "<h1>API for climatemind ontology</h1>", 200
 
 
-@app.route("/signup", methods=["POST"])
+@app.route("/subscribe", methods=["POST"])
 @auto.doc()
-def signup() -> Tuple[Response, int]:
-    """
-    Users want to be able to sign up with their email to save their personal values test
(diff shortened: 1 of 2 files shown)


y-himanen:
this won't work with what I've written because the store_subscription_data function needs to be edited too. let's chat and figure out the best way to do this.

knowledge_graph/routes.py
```diff
@@ -314,64 +278,70 @@ def get_feed():
...
-    recommended_general_myths = MYTH_PROCESSOR.get_user_general_myth_nodes()
-    climate_general_myths = {"myths": recommended_general_myths}
-    return jsonify(climate_general_myths), 200
+    try:
+        recommended_general_myths = MYTH_PROCESSOR.get_user_general_myth_nodes()
+        climate_general_myths = {"myths": recommended_general_myths}
+        return jsonify(climate_general_myths), 200
+    except:
+        error = {"error": "Failed to Process Myths"}
+        return make_response(error), 400
```
NickCallaghan, line 333: We should return a 500 here as its a server problem

NickCallaghan (approved):

y-himanen:
do not merge this, please

y-himanen:
changes will be incorporated elsewhere

PR #166, opened 2021-01-21
general solutions order changed - from highest to lowest co2 equivale…

y-himanen:
…nt reduced values from project drawdown


rodriguesk (approved):
Beautiful!

PR #173, opened 2021-01-24
added /post-code endpoint

seanmajorpayne:
Added /post-code endpoint which accepts a Post request with a sessionId and postCode. Then stores that postCode to the Sessions Object related to the particular sessionsId.

Removed zip code processing from the scores endpoint.

Added nuanced error messages.


seanmajorpayne:
On Nick's request, I have added back in the post code to the scores endpoint so the front-end will not break. We can always remove this later.

rodriguesk (approved):
I didn’t test this, but I reviewed the code and looks good.

PR #179, opened 2021-02-03
edited migration docs, corrected typo in readme, edited entrypoint.sh…

y-himanen:
… for clarity

### DIFF
--- a/docs/MIGRATIONS.md
+++ b/docs/MIGRATIONS.md
@@ -1,24 +1,31 @@
 # Migrations
 
 When changes are made to the database structure, migrations are performed with Flask-Migrate to
-update these changes within SQLalchemy.
+update these changes within SQLAlchemy. Our migration scripts are auto-generated using Alembic.
 
-Migrations happen in two places, locally and within the Azure Cloud. The steps to run the migrations
-differ slightly between the two.
+Migrations happen in two places, locally and within the Azure Cloud, where our production database is hosted. The steps to run these migrations
+are different. Please read the instructions carefully and contact the team if you have any questions.
 
-You will need to run a migration anytime the models in the models.py file are changed.
+You will need to run migrations any time the models in the models.py file are changed. A local migration should be completed before any changes are made in the cloud.
 
-## Running A Local Migration
+We have chosen our database tools to make it as easy as possible for developers to interact with the database. However, it is important to understand a little bit about what is happening in the background. 
+
+Looking at entrypoint.sh you will see that there are two commands that are run every time a migration is done - the migrate command and the upgrade command. The migrate command is what triggers the auto-generation of the migration script. Alembic will look for any changes in the models and automatically create a migration script based on these. These changes are then applied through the upgrade command.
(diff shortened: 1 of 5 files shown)


docs/MIGRATIONS.md
```diff
@@ -1,25 +1,32 @@
...
 
-## Running A Local Migration
+We have chosen our database tools to make it as easy as possible for developers to interact with the database. However, it is important to understand a little bit about what is happening in the background. 
+
+Looking at entrypoint.sh you will see that there are two commands that are run every time a migration is done - the migrate command and the upgrade command. The migrate command is what triggers the auto-generation of the migration script. Alembic will look for any changes in the models and automatically create a migration script based on these. These changes are then applied through the upgrade command.
+
+For local migrations, the migrate and upgrade commands are run one after another. In practice, this means that Docker will generate an empty set of tables every time the project is spun up locally, with the exception of the LRF (location relevance flag) data. The LRF table will be generated with a full set of data every time, using a separate script that is also run as part of the local migration process.
```
rodriguesk, line 15: “In practice, this means that Docker will generate an empty set of tables every time the project is spun up locally, with the exception of the LRF (location relevance flag) data. “
This sentence makes it sound like the migrate and upgrade commands are what make the local database empty each time. Which isn’t true. It’s the fact that docker-compose.yml is just used for local docker containers to spin up, while in the cloud there’s a different process (just think of it like the cloud uses a different docker-compose.yml.) And the local docker-compose.yml pulls a fresh blank Microsoft sql database docker image from online. Whereas in the cloud docker container startup it just connects to the cloud database. In practice this means that the local database is blank every time the local docker containers are spun up. So a first time migration must occur every time the docker containers are spun up locally.

And because there is LRF data required for the app to run, this data is loaded into the database in a separate step that occurs after migrations are completed through the entrypoint.sh

Should say something like what I wrote above.

rodriguesk (approved):
We will eventually want the Azure cloud migration instruction not to have anyone add any passwords or anything to the entrypoint.sh file as I really worry someone will accidentally commit azure passwords on accident.

I think what’s missing is that there’s an azure IP address firewall for our cloud database so IP addresses need to be added to our firewall to get access to the cloud database for cloud migrations.

We should probably switch to have a secret file that is git.ignored and that gets referenced for azure database params credentials instead of them being typed in manually by people doing cloud migrations? That might be a safer way that way credentials don’t get accidentally committed to GitHub.  And also migrations for cloud could also be forced to be done through an interactive commands entered manually through commend prompt in interactive mode instead of through saving it in the entrypoint.sh script and then having it get run upon app startup. 

This might need to be noted that we intend to change the process soon to be like this above for cloud migrations to enhance security for climatemind.

docs/MIGRATIONS.md
y-himanen: I provided a simpler explanation. I don't think it's worth pulling info on Docker images into this part. Merging this now. Let's continue our discussion on the db params separately.

PR #180, opened 2021-02-03
Cm 478 test myths

seanmajorpayne:
@NickCallaghan Any feedback on this is appreciated. I just tried to use a similar format to you, but if there is more I need to add please let me know.


NickCallaghan (approved):
Nice Work!! Might be nice to check the data type for each myth property. Front end tends not to like it expects one data type and  gets another.  

A basic check for a list/object/string would be handy but up to you if you want to do now or later. :)

seanmajorpayne:
@NickCallaghan Thanks for the prompt review! I'm adding in a check for the data type tonight and then I'll merge.

PR #181, opened 2021-02-05
Cm 480 solutions test

seanmajorpayne:
Basic solutions test now working, includes type verification.


seanmajorpayne:
Merging without review since this will not break production or affect production code.

PR #185, opened 2021-02-12
updated docker-compose to include cloud_db_params variable, updated m…

y-himanen:
…igrations doc, tested connections & migrations still work (ran migrate without upgrade for cloud db to check alembic tracking)


rodriguesk (approved):
I still think it’s best to name the file secret.env so it’s very clear it’s secret

y-himanen:
It's done un-named so that it will be automatically detected. It can also have a name, but we would have to declare it in a different way in the docker-compose.yml. We would have to drop the environment: tag and use env_file: to make it work, and it seems like you can't have both in the same compose file. So given that we have other things listed under environment in addition to the db_params, it makes sense to do it this way. Let me know if you have questions!

PR #206, opened 2021-04-28
CM-624_fix_for_all_same_values

NickCallaghan:
# Backend fix for scores normalisation issue

## Issue
When the user answers the same for every question the normalisaton function was throwing a divide by zero error

## Fix 
- Added contitional check where if the max - min values == 0 then a score in the middle of the normalised range is returned for each Personal Value
- Added cypress test to cover this test case.

### DIFF
--- a/cypress/integration/scores.spec.js
+++ b/cypress/integration/scores.spec.js
@@ -1,6 +1,7 @@
 /// <reference types="cypress" />
 
 import scores from "../fixtures/postScores.json";
+import scoresAllTheSame from "../fixtures/postScoresAllSameAnswer.json";
 import scoresSetTwo from "../fixtures/postScoresSetTwo.json";
 
 let sessionId; // store the session id from the post to chech we can retreive the values and scores.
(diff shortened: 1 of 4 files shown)


cypress/integration/scores.spec.js
```diff
@@ -86,4 +87,24 @@ describe("Scores endpoint", () => {
...
   });
+
+  it("When the user answers the same for every question each value should be scored in the middle of the range", () => {
+    //Post scores with all the same answer
+    cy.request("POST", "/scores", scoresAllTheSame).should((response) => {
+      sessionId = response.body["sessionId"];
+
+      // Check the personal values scores for the corresponding sessions are correct
+      cy.request(
+        `http://localhost:5000/personal_values?session-id=${sessionId}`
```
ManilTester, line 98: Can we make the port number a configurable variable from cypress.json

NickCallaghan:
> Can we make the port number a configurable variable from cypress.json. 

I've removed the host url altogether, it's already defined in the cypress.json file

rodriguesk (approved):
need to fix linting, but other than that all good from my end.

PR #208, opened 2021-05-03
CM-563 Add register endpoint

seanmajorpayne:
Input validation for full_name, email, and password
Detailed errors for each negative circumstance
Working security endpoints for registration and login
Modified local and cloud DB
Rewrote Users table slightly


y-himanen (approved):
I pulled and tested and everything works great! A few things that I think could be addressed:

- I would like to see a different error message for when the registration failure is related to the email or the password. I don't like either or messages that leave you guessing which one went wrong. Also, @NickCallaghan - I hope that frontend will provide the user with some explanation of the password requirements? I hate when a website makes me guess why my password is failing!!
- there's no check that a full name is actually included - is this important? checking for 2-50 characters will allow a name like "dou" to be registered. We also might want to consider some including some whitespace stripping.
- there's a couple of typos/missing words in the error messages, e.g. auth/routes.py line 40

seanmajorpayne:
Thanks for reviewing!

The errors are just what Nick had requested. Nick do you not want something
more nuanced?

I'll fix the typos. Full name is probably being replace with separate first
name/last name so I didn't make a check for it as of now.

seanmajorpayne:
Also there are a couple of places we don't want to do nuanced errors for
security reasons. We don't want to confirm whether or not it's a bad email
or password for login for example to provide minimal feedback for bad
actors.

PR #209, opened 2021-05-08
Cm 662 alter solutions

seanmajorpayne:
Added political filtering for solution nodes.

Session-id is an optional parameter for the /solutions endpoint. If a session-id is provided, the user is run through the ML model to predict whether or not they are radically political on either spectrum. Then this is cross referenced with the recently added solution node labels to determine whether or not the node conflicts with the user.

If session-id is not provided, the endpoint functions normally returning all solution nodes.


NickCallaghan:
Can you add a cypress test covering what you have added?  :)

app/solutions/routes.py
```diff
@@ -37,13 +41,84 @@ def get_general_solutions():
...
         return jsonify(climate_general_solutions), 200
+
     except:
         raise CustomError(
             message="An error occurred while processing the user's general solution nodes."
         )
+
+
+def predict_radical_political(user_scores):
+    """
```
AlexisJCarras, line 71: Models have been trained on data that has been augmented in different ways. For example, our current best performing model was trained on Ranked data (i.e., data containing the binary pairwise comparisons between the 10 different attributes). Perhaps we should think about implementing those processing steps here, to be applied to user_scores.
If the data is not correctly process to match the training data of the loaded model, .predict() would throw an "operands could not be broadcast together with shapes {}{}" error.

app/solutions/routes.py
```diff
@@ -37,13 +41,84 @@ def get_general_solutions():
...
+    if user_scores:
+        user_liberal, user_conservative = predict_radical_political(user_scores)
+    else:
+        user_liberal, user_conservative = None, None
+
     try:
         recommended_general_solutions = (
-            SOLUTION_PROCESSOR.get_user_general_solution_nodes()
+            SOLUTION_PROCESSOR.get_user_general_solution_nodes(
+                user_liberal, user_conservative
```
AlexisJCarras, line 58: As a next step, to make the integration we can think of ways to deal with conflicting predictions (e.g., a case where user_liberal ==1  && user_conservative==1).

AlexisJCarras (commented):
Everything looks great for the current ticket.
As next steps:
1. We will have to make sure that user_scores is processed in the exact way that the model expects the data (some models have been trained on additionally processed data). For now, given that the two random forest models were trained on the scored 10 personal values the implementation is fine.
2. We can think about smarter 'cooperation' mechanisms between the two classifiers, to deal with conflicts in predictions, for example.

y-himanen (commented):
The happy path works great!

But if there is any issue with processing through the ml model, the app will crash (I crashed it intentionally with Postman a couple of different ways). I suggest that some exception handling be added to deal with scenarios where there are issues with processing through the models, or if an incorrect/malformed session-id is passed etc. 

So app/solutions/process_solutions.py line 173 `get_scores_vector `needs to account for the db query failing. And line 70 `predict_radical_political` needs to account for pickle failure.

Also, lint failed on this round, so automated tests aren't currently passing.

seanmajorpayne:
@y-himanen Thanks! I've added in error checks based on those situations.

y-himanen (approved):
great work!!

PR #211, opened 2021-05-25
CM-700-quiz-retake

seanmajorpayne:
Workflow has been updated for the scoring endpoint

Old way:
1) Non-registered user takes the quiz, then registers and this links the current session of scores to their account.

New Way:
1) User can take the quiz when already registered
2) Scores endpoint will link the session to the user by JWT token
3) If no JWT token is present, the old way will still be used
4) Frontend can obtain the new session id on login

Soon there will be no reason to return the session-id on login.


y-himanen (approved):
Great work!! Review process comments added to ticket as agreed.

PR #212, opened 2021-05-25
CM-569_BE_user_can_logout



NickCallaghan:
This PR is required as part of the logout story to unset the users refresh token, in effect loging them out the app. 

- Add /logout endpoint to api which unsets the users refresh token.

- Add basic cypress test for logout

- Fixed some unrelated tests which had been broken some how.  had to comment out some of the sad path tests for  register, the api is throwing a 500 error but it's unrelated to this work. 
- Some of the tests are not very nice as we have to make multiple calls to get the api in the correct state to carry out the test. This is why we need data seeding.

- Unfortunately it's also necessary to remove the following records from the user table before each test run locally. Again, this is why we need data seeding.
```
DELETE from users where email='test1@example.com'
DELETE from users where email='test@example.com'
DELETE from users where email='login@example.com'
```


NickCallaghan:
# User can log out CM-569

created logout endpoint which unsets refresh token.

Added cypress test for logout

seanmajorpayne:
Hold off on merging. I need to double check if this will actually invalidate. I'm concerned a user will be able to circumnavigate this.

NickCallaghan:
> Hold off on merging. I need to double check if this will actually invalidate. I'm concerned a user will be able to circumnavigate this.

It won't invalidate the token it will just unset it in the client so that it no longer has it to use. We can do token blacklisting but i'd be inclined not to over complicate things for the first pass.

seanmajorpayne:
I've updated logout to use the built in jwt-extended function for unsetting cookies.

I've also removed the users endpoint as it shouldn't be included in this PR since we've deprecated the need for it.

This isn't passing cypress tests, so once those are updated feel free to merge. Thanks Nick!

seanmajorpayne (approved):

PR #213, opened 2021-05-30
processed new ontology file 30 may

y-himanen:


seanmajorpayne:
Hey Yasmine, Thank you for all of the hard work last night.

I believe the old ontology file was one I put up using my local instance of the ontology refactor, which is why code broke when you updated the ontology.

I've met you in the middle here. There is no need to check if cons/lib are in the data properties. They will always be there as the ontology processing will add it even if the result is "Conservative": None. None defaults to False in Python. So two lines is enough to handle this whole code. I've just adjusted to match the old format with ["data_properties"] included in front of the labels.

Sorry for the confusion. That is completely my bad. I've updated my local ontology PR branch to match this format, so when that goes live, there won't be any breaking changes.

seanmajorpayne (approved):

PR #217, opened 2021-05-31
added uuid validation for solution get request

y-himanen:


app/solutions/routes.py
```diff
@@ -128,7 +141,7 @@ def get_scores_vector(session_id):
                 scores.universalism,
             ]
         else:
-            return None
+            return "Not in db"
```
seanmajorpayne, line 144: Looks good overall. Thanks for knocking this out.

You can throw the InvalidUsageError here instead of returning. Although Python allows us to assign a variable to multiple types, we ideally want to use each variable for one type.

seanmajorpayne (changes requested):

app/solutions/routes.py
y-himanen: I may have already commented, but I think either I didn't save it correctly, or it's squirreled away somewhere.

Closed this because I had already tried it that way and couldn't get it to extricate the error message from the one in the original try block. You're welcome to try and come up with another solution, but closing this as fixed as for now.

seanmajorpayne:
No problem. It works and it's not urgent to fix.

PR #222, opened 2021-06-05
CM-701

econaxis:
Implemented CM-701. Added separate error messages via `InvalidUsageError` for each field that could be wrong (email and password). Reiterated password requirements when password is wrong, reiterated their provided email when email is wrong.

Tested locally on my machine by submitting bogus emails/passwords and checking that the error messages are returned.


seanmajorpayne (approved):

seanmajorpayne:
Nice work! Looks great.

PR #225, opened 2021-06-13
CM-568_refresh_access_token

NickCallaghan:
I’ve managed to get the refresh token working. There were a couple of issues.

Backend: In order to set the cookie the cors header cannot be *. 

Frontend: The front end needs to set {withCredentials: true} on both the /login and /register calls to allow the backend to set the cookie.

Access token: should be valid for 15 minutes, i've updated this


seanmajorpayne (approved):
Looks great Nick. Thank you!

PR #234, opened 2021-07-09
Cm 740 personal values endpoint

y-himanen:


app/personal_values/routes.py
```diff
@@ -17,17 +17,17 @@
 def get_personal_values():
     """
     Users want to know their personal values based on their Schwartz questionnaire
-    results. This returns the top 3 personal values with descriptions plus all scores for a user given a session ID.
+    results. This returns the top 3 personal values with descriptions plus all scores for a user given a quiz ID.
     """
     try:
-        session_uuid = uuid.UUID(request.args.get("session-id"))
+        quiz_uuid = uuid.UUID(request.args.get("quizId"))
```
rodriguesk, line 23: Just wondering why previous version had a dash both now this quizId doesn’t.

app/scoring/process_scores.py
```diff
@@ -87,11 +87,8 @@ def persist_scores(self, user_uuid):
 
         """
         try:
-            user_session = Sessions(session_uuid=self.value_scores["session-id"])
-            db.session.add(user_session)
-
             user_scores = Scores()
-            user_scores.session_uuid = self.value_scores["session-id"]
+            user_scores.quiz_uuid = self.value_scores["quiz_uuid"]
```
rodriguesk, line 91: Just pointing out again that dash has been dropped and differently this time too.

rodriguesk (approved):
Looks good. Just recommending double checking that transition away from session-id is correct because in some places it changed to quizId while other places it changed to quiz_uuid.

app/scoring/process_scores.py
y-himanen: Dropping dashes is just making them consistent. Whenever we're talking about the quiz id from the backend perspective it's quiz_uuid and whenever it's coming from the frontend (e.g. as part of request body) then it's quizId. This is in keeping with style differences between JS and Python. And the quizId is the agreed name with frontend. Have postman tested, with edited request parameters to reflect the change. FYI @NickCallaghan

PR #235, opened 2021-07-09
Cm 741 feed

y-himanen:


app/scoring/routes.py
```diff
@@ -78,23 +77,13 @@ def user_scores():
...
+    quiz_uuid = uuid.uuid4()
+    value_scores["quiz_uuid"] = quiz_uuid
 
     user_uuid = None
     if current_user:
-        user_uuid = current_user.uuid
+        user_uuid = current_user.user_uuid
     process_scores.persist_scores(user_uuid)
 
-    postal_code = parameter["zipCode"]
```
rodriguesk, line 89: The frontend was supposed to switch over to have postal code be a separate endpoint than questionnaire scoring. Did frontend never switch over as they were supposed to? It doesn’t make a lot of sense to have postal code be associated with the questionnaire. Sean made this other endpoint and it should be functional. What’s the issue with it?

rodriguesk (approved):
Just want to know why we need to delete the endpoint that the frontend is supposed to transition to use. Sean set that up a while ago and frontend just hasn’t transitioned yet to use it (unless I’m getting the two endpoints mixed up).

app/scoring/routes.py
y-himanen: This one was a bit confusing to follow because I went back and made changes to CM-743 and then merged them into this one to keep the code cumulative. But I didn't delete the postcode endpoint. I removed any postcode processing from the scores endpoint to complete the switchover to using the new dedicated endpoint.

PR #237, opened 2021-07-09
auth endpoint updated + docstrings requested for session_helpers adde…

y-himanen:
…d. more details in jira ticket

### DIFF
--- a/app/auth/routes.py
+++ b/app/auth/routes.py
@@ -153,7 +153,7 @@ def register():
 
     if not r:
         raise InvalidUsageError(
-            message="Email and password must included in the request body"
+            message="Email and password must be included in the request body."
         )
 
     first_name = r.get("firstName", None)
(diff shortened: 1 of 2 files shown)


app/auth/routes.py
```diff
@@ -155,23 +156,25 @@ def register():
...
-    full_name = r.get("fullname", None)
+    first_name = r.get("firstName", None)
+    last_name = r.get("lastName", None)
     email = r.get("email", None)
     password = r.get("password", None)
-    session_id = r.get("sessionId", None)
+    quiz_uuid = r.get("quizId", None)
 
-    if not valid_name(full_name):
+    if not valid_name(first_name) or not valid_name(last_name):
```
rodriguesk, line 165: I think the errors should be separate for the first and last name, unless the front end is also requiring form validation so the user knows their first name is the problem or their last name is the problem. If the user is just told either their first name or their last name (but not which one exactly) is causing problems, that is likely to frustrate them.

app/session/session_helpers.py
```diff
@@ -22,7 +37,7 @@ def store_session(session_uuid, session_created_timestamp, user_uuid):
 
 def process_ip_address(request, session_uuid):
     """
-    Save a user's IP address information into the database with their session_uuid.
+    Check's the user's IP address information and calls another functoin to store it in the database with their session_uuid.
```
rodriguesk, line 40: Type in the word function

app/session/session_helpers.py
```diff
@@ -57,6 +72,16 @@ def process_ip_address(request, session_uuid):
 
 
 def store_ip_address(ip_address, session_uuid):
+    """
```
rodriguesk, line 75: Thank you for adding this docstring! I owe you as I’m pretty sure this was my fault for not including when I submitting this code originally l!

rodriguesk (approved):
Looks awesome! Great work. Just few tiny things but nothing major.

PR #241, opened 2021-08-01
updated subscribe, feed and scores endpoints to put session id in tables

y-himanen:


app/scoring/routes.py
```diff
@@ -83,7 +84,27 @@ def user_scores():
     user_uuid = None
     if current_user:
         user_uuid = current_user.user_uuid
-    process_scores.persist_scores(user_uuid)
+
+    session_uuid = request.headers.get("X-Session-Id")
+
+    if not session_uuid:
```
seanmajorpayne, line 90: Okay, I see what you mean now about the duplicate session_uuid processing.

I think it's worth pulling this out into a separate function in this PR since it should be a very quick improvement.

Everything from "if not session_uuid:" until "valid_session_uuid" can be ripped out and put into one function to replace these calls.

So end result would be
```
session_uuid = request.headers.get("X-Session-Id")
valid_session_uuid = validate_uuid(session_uuid)
```

Keeping the generic validate_uuid so that this makes sense for more than just sessions as we do this in a number of places with the user_uuid and quiz_uuid as well.

Especially if we will need to do this on all endpoints from now on, I'd rather have us update sooner than later.

seanmajorpayne (changes requested):
Great job with this @y-himanen and @SabineU.

I'd recommend pulling that code out and then it's good to go.

seanmajorpayne (approved):

app/scoring/routes.py
y-himanen: I think we will put this improvement into a separate ticket, as it does not impact on functionality. Thanks for your review.

SabineU (approved):

PR #244, opened 2021-08-05
CM-801 cleaning up auth endpoints

seanmajorpayne:
- Removed a bunch of "if not" statements and concatenated them into one for loop
- Removed helper methods and made them inline since most are used in one place
- Fixed small bug where if/elif statement was used when both first/last name need to be validated
- Removed unnecessary return true/false and replaced with returning objects/None
- Removed unnecessary password format validation from login (if it's not right, it's not right, we don't care if it's formatted correctly)
- Removed email format check from login (SQL Injection is not possible with filter_by, only with filter, so we don't care if the format is wrong)

Most of these methods you can now read top to bottom without needing to jump constantly to somewhere else in the code. And the code is slightly faster.


app/auth/routes.py
```diff
@@ -153,52 +153,54 @@ def register():
 
     if not r:
         raise InvalidUsageError(
-            message="Email and password must be included in the request body."
+            message="Parameters must be included in the request body."
```
econaxis, line 156: I think the previous one was a more descriptive error?

app/auth/routes.py
```diff
@@ -153,52 +153,54 @@ def register():
...
 
-    if not scores_in_db(quiz_uuid):
+    scores = db.session.query(Scores).filter_by(quiz_uuid=quiz_uuid).first()
+
+    if not scores:
         raise DatabaseError(message="Quiz ID is not in the db.")
 
-    if not check_email(email):
+    if not check_email(r["email"]):
         raise InvalidUsageError(message=f"The email {email} is invalid.")
```
econaxis, line 188: {email} is not defined (deleted above).

econaxis (commented):
Just one error but everything else looks good.

app/auth/routes.py
seanmajorpayne: It was, but there are only 2 parameters and we have 5 of them total that should be included. Do you think we should specify all 5?

app/auth/routes.py
seanmajorpayne: Good catch. Thanks!

app/auth/routes.py
seanmajorpayne: Actually, I think it makes the most sense to say JSON is required, as this is not parameter specific but just whether or not JSON is included and the frontend team is the only group accessing this endpoint. Individual missing params will get caught in the forloop.

y-himanen:
@seanmajorpayne I have sent you comments through Slack. Please make edits, and I will quickly recheck.

PR #247, opened 2021-08-09
CM-832 conversations new

seanmajorpayne:
Added endpoints to create a conversation and to query all existing conversations for a user.

X-Session-Id and Authorization headers are both required to access these endpoints.

@y-himanen Since I needed to validate the session-id I went ahead and added a utils file to our Auth. I copied your code style for the session checks into here and used an Enum so we can make it generic and use it for every type of UUID in the app. Enums are fast since they are at the byte level of the computer. In the near future on your proposed ticket, we can simplify the other UUID checks to call this function.


seanmajorpayne:
Update CM-808 has been added along with cleaning up some other small bits of code.

NickCallaghan (approved):

NickCallaghan:
@seanmajorpayne Pushed a couple of minor changes: 
 - Can we start the status enum from 0 not 1. So status 0 = Invited
 - Update the route names GET /conversations and POST /conversation. We'll be wanting PUT, GET,DELETE endpoint on conversation for the other CRUD operations so /create-a-conversation doesn't make sense. 

Front end is now in PR for the first bit of conversations so would be good to get this merged

NickCallaghan:
We should strongly consider making X-Session-Id not required on any endpoint. It creates a single point of failure so if for some reason the backend can't issue a new session id or the front end gets our of sync then pretty much the whole app fails.  :( 

Given it's not needed for functionality and only analytics would it not make sense to allow the request (and maybe log it somewhere). It's had a few instances although mainly down to the DB being missing the session id. Wonder if we might see it in live.

rodriguesk:
Tracking sessions is a requirement.

There should be no cases where it gets out of sync. And no cases where backend can’t issue a new sessionid. 

Basically, the app isn’t designed to work without internet.

If we want to make the app work without internet, then we can think of that at a later point.

NickCallaghan:
> Tracking sessions is a requirement. There should be no cases where it gets out of sync. And no cases where backend can’t issue a new sessionid. Basically, the app isn’t designed to work without internet. If we want to make the app work without internet, then we can think of that at a later point. 

Shouldn't be and won't be are 2 different things.  As it stands if we break the session endpoint we break the whole app. 

We should at the very least be logging these sort of things so we can see if there are requests being refused  / failing

NickCallaghan:
> Basically, the app isn’t designed to work without internet.

I was more talking about if we break that App, thats probably more likely

rodriguesk:
Is this idea the same as CM-758? It’s currently in PR.

NickCallaghan:
CM-758 is DONE https://climatemind.atlassian.net/browse/CM-758

so assume you mean a different ticket. not sure what one.

rodriguesk:
https://climatemind.atlassian.net/browse/CM-759
Sorry, meant CM-759

NickCallaghan:
Was meaning logging from production and test sites. That is for
local logging to a file for debugging.

Not sure what we are presently logging (if anything) backend wise. But all
500s, crashes, refused requests would be a good starting point.

Once we are logging it, how is it being monitored? Whos responsible for
flagging up issues.

rodriguesk:
That would probably be me in terms of responsible for checking things. We will need to get email notifications working for users soon and when that happens we can set it up to also email me or some error email to make error tracking more accessible for everyone on the team.

SabineU (dismissed):

seanmajorpayne:
Updated. Ready for final review.

SabineU (approved):

y-himanen (approved):

PR #248, opened 2021-08-14
Cm 813 rate limiting

seanmajorpayne:
Adding (somewhat) strict rate limiting for security points.
General rate limits
and no limit for refresh tokens.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from datetime import timezone
 from datetime import timedelta
 from flask_cors import CORS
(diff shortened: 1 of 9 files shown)


y-himanen (approved):
This looks good for now. We may need to adjust the rate limits later on. 

Also, @NickCallaghan and @SabineU - can your user generation code for the cypress tests be aligned somehow? Nick has added a library for this job, where Sabine added a random user generator to utils for a different ticket. Would be good if we picked one or the other.

SabineU (dismissed):

PR #255, opened 2021-08-31
user can update email - new email endpoint

y-himanen:


SabineU (approved):

app/email/routes.py
```diff
@@ -0,0 +1,90 @@
...
+        raise InvalidUsageError(
+            message="Cannot update email. Email is not formatted correctly."
+        )
+
+    if new_email != confirm_email:
+        raise InvalidUsageError(
+            message="Cannot update email. New email address and confirm new email address do not match."
+        )
+
+    email_already_used = Users.query.filter_by(user_email=new_email).first()
```
seanmajorpayne, line 71: I think it makes sense here just to leave this as user = Users.query...

As you know, the object returned is a user. It's easy to understand
```
if user:
    raise AlreadyExistsError(message="A user is already registered with this email.")
```

In fact, in the registration endpoint I used this format (Note that we have a method built in the user class to handle this query):

```
    user = Users.find_by_username(r["email"])
    if user:
        raise UnauthorizedError(message="Email already registered")
    else:
        user = add_user_to_db(
            r["firstName"], r["lastName"], r["email"], r["password"], r["quizId"]
        )
```

It might make sense to rename the find_by_username to find_by_email, but it doesn't matter much to me since they are equivalent at the moment, so I'll leave that up to your choice.

Also, I think you meant to return the email they typed? Right now you are returning a literal message "Cannot update email. Email"

Additionally, this is a place we should be using one_or_none instead of first since only one user should exist per email. If you use the find_by_username function, then this is already set up this way.

app/email/routes.py
```diff
@@ -0,0 +1,90 @@
...
+    if email_already_used:
+        raise AlreadyExistsError(message="Cannot update email. Email")
+
+    if not current_user.check_password(password):
+        raise UnauthorizedError(message="Cannot update email. Incorrect password.")
+
+    try:
+        current_user.user_email = new_email
+        db.session.commit()
+    except:
```
seanmajorpayne, line 83: You can use
```
from sqlalchemy.exc import SQLAlchemyError
...
...
except SQLAlchemyError
```

I'm hoping we can be more explicit about catching errors so when we get logging set up, we'll know if unexpected behavior is happening with any of the error catching.

seanmajorpayne (changes requested):
Looks good! A couple of suggestions for updates.

seanmajorpayne (approved):

app/email/routes.py
y-himanen: Discussed and agreed alignment strategy to use class method and AlreadyExistsError.

app/email/routes.py
y-himanen: Done.

seanmajorpayne:
I'm going to take a closer look at this issue today after my interviews.

I'm thinking it might have to do with this app_context thing since the blueprints are declared within the context, but the db is not created in that context. There could be some really weird session management issues with that.

seanmajorpayne:
Sure, makes sense to me. We can switch to a more robust library.

That regex is not something I custom wrote. It is the same check you'd find
in a library. It's the RFC standard for emails.

PR #263, opened 2021-10-10
refactoring scores

### LINKED ISSUE #267: Feed & Score Nodes Refactor
Issues exist in feed.py and score_nodes.py and should be refactored. Score nodes presently uses unconventional python naming with variables in ALL_CAPS format, has a few variables that are declared and never used, and does not use dependency injection as it should. It is only used by the feed, but is located in the scoring folder so it should be relocated. Imports should be in Pep8 order & relative imports should be used when possible. Storing the climate feed into the db should not occur in this class.

Feed uses some variables in ALL_CAPS, a generalized try/except to catch multiple types of errors, and imports are not in pep8 order.Both should use Pep 287 reST formatting for docstrings.

We're passing a quiz_uuid to a query for a session_uuid in the localized graph function and this is bad!

seanmajorpayne:
Fixes #267 

- Refactored for Pep-8 & Pep-287
- Updated imports for Pep-8 (Standard Library first, 3rd party second, our app 3rd & Alphabetical for each section)
- Replaced ALL_CAPS variables
- Switched unnecessary self variables to function variables
- (Fixed?) Changed localized graph to actually query the session_id instead of giving the Sessions object a quiz_uuid to query by
- Moved test ontology creation outside of the feed & into the app config (this doesn't need to be done individually per user)
- Added dependency injection for the class instead of declaring other classes inside
- (Probably?) Significantly faster since this removes a pass through the ontology. Need to see test results.
- Removed num_feed_cards variable that was being passed through 3 functions and added it directly to function where used

In a separate PR, we need to figure out if the localized graph is doing what it is supposed to do.

### DIFF
--- a/app/feed/get_feed.py
+++ b/app/feed/get_feed.py
@@ -10,7 +10,7 @@ import numpy as np
 from ..scoring.build_localised_acyclic_graph import build_localised_acyclic_graph
 
 
-class GetFeed:
+class Feed:
 
     """
     After users have completed their survey, they want to receive personalized suggestions
--- a/app/feed/routes.py
+++ b/app/feed/routes.py
@@ -6,7 +6,7 @@ from sqlalchemy.exc import SQLAlchemyError
 
 from app import auto, cache, db
 from . import bp
-from .get_feed import GetFeed
+from .get_feed import Feed
 from .store_climate_feed_data import store_climate_feed_data
 from ..auth.utils import check_uuid_in_db, uuidType, validate_uuid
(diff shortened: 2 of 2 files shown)


seanmajorpayne:
Fixes #265

seanmajorpayne:
Fixes #265

seanmajorpayne:
fixes #265

seanmajorpayne:
@rodriguesk Reverted feed cards based on your feedback. Ready for review.

PR #268, opened 2021-10-10
CM-817 Building the scaffold to run pytest

Moataz-Mahmoud:


rodriguesk:
@Moataz-Mahmoud you can merge this into develop now.

PR #270, opened 2021-10-20
CM-864 Email Welcome

seanmajorpayne:
Sends welcome email to address. Testing at the moment. Please don't review/merge.


rodriguesk:
@seanmajorpayne yes please use the sendgrid API (instead of the SMTP relay). That’s what I was hoping you would do.

We can have 1000 emails a month, so it should be good to set up for the error reporting.

User should be able to register even if the welcome email fails. It’s intended to be welcome email to get you familiar with the sendgrid API. The story is not meant to be a confirmation email nor a validation email. Let’s try not to over complicate it. It’s ok to add new stories to the backlog though for later if you guys decide that’s needed.

y-himanen:
Closed as out of date.

PR #271, opened 2021-10-30
CM-873 User loses feed with refresh

seanmajorpayne:
Tested with an icognito window in Google Chrome and confirmed behavior is fixed locally.


seanmajorpayne:
@johols Please confirm this works for you.

@y-himanen @SabineU The two email tests for the expired Token have strange behavior. Locally, the endpoint returns a 422 error for an Invalid Token. However on CircleCI the test returns a 401 code for an Expired Token as expected.

Both errors are fine as far as I'm concerned, but it would be nice if the test could somehow pass in both enviornments. Any idea why it would differ between these two environments?

PR #272, opened 2021-11-06
updated set 1 questions with shorter/simpler text

y-himanen:


y-himanen:
@rodriguesk - If you're happy with the changes then I think you can just merge this. The backend testing doesn't need updating in any way for this.

PR #276, opened 2021-12-05
Cm-969 changed name length requirements for registering users and for inviti…

y-himanen:
…ng user b - name <=20. error messages updated


rodriguesk (approved):
Nice work @y-himanen !

PR #278, opened 2021-12-30
get single conversation endpoint created

y-himanen:


y-himanen:
image

NickCallaghan:
Thanks for getting this up @y-himanen!

would it be possible to have the path as `conversations/<conversationId>` rather than the conversationId as a param? works ok as is but not really the best practice in terms of naming. 

Might also want to consider nesting the user a and b stuff inside an object in case we add more stuff later:

``` 
userA : { 
              'name' : 'Nick',
              'id' : 123
         }
```

y-himanen:
image

- no probs to change. followed the 'way we did it before' based on our earlier conversation.

image

- 404 will come back on all endpoints where something doesn't exist now, rather than the catch-all 500 we were using.

NickCallaghan (dismissed):

app/conversations/routes.py
```diff
@@ -145,3 +144,23 @@ def get_conversations():
     response = {"conversations": results}
 
     return jsonify(response), 200
+
+
+@bp.route("/conversation/<conversation_uuid>", methods=["GET"])
+@cross_origin()
+def get_conversation(conversation_uuid):
+    """
+    Get a single conversation.
```
rodriguesk, line 153: This needs a parameters section in the description and a description of the response.

app/conversations/routes.py
```diff
@@ -145,3 +144,23 @@ def get_conversations():
...
+    """
+    session_uuid = request.headers.get("X-Session-Id")
+    session_uuid = validate_uuid(session_uuid, uuidType.SESSION)
+    check_uuid_in_db(session_uuid, uuidType.SESSION)
+
+    validate_uuid(conversation_uuid, uuidType.CONVERSATION)
+    check_uuid_in_db(conversation_uuid, uuidType.CONVERSATION)
+    response = build_single_conversation_response(conversation_uuid)
```
rodriguesk, line 161: I think it might be better to not have a helper function build the response unless this is done with the other endpoints. The reason I say this is because if we want to edit the response it would be good to have all the response code in the same place as the other endpoints. It also makes documenting the endpoint easier because wouldn’t have to change the documentation in multiple places.

app/conversations/routes.py
```diff
@@ -145,3 +144,23 @@ def get_conversations():
...
+
+    start_user_b_journey(conversation_uuid)
```
rodriguesk, line 163: I think this is a smooth way to have user b events logged for analytics. But in this specific case I think this endpoint is used to retrieve the conversation card information to populate for user a? If so I don’t think this is a user B journey event and is a user A journey event. User A journey events are probably best to be tracked by the frontend since it’s already been tracking user A events with google analytics.

Also, when we want to add new analytic events, before it gets coded up we should add the new analytic event type and it’s description (whether frontend or backend) to the analytics page on Confluence so analytics people know what the event type means (they can read a description for each event type when it appears in the analytics data). This also helps to ensure we all understand and on same page for what a specific event is and when it should be triggered.

app/conversations/utils.py
```diff
@@ -0,0 +1,35 @@
+from app.models import Conversations, Users
+from app import db
+
+
+def build_single_conversation_response(conversation_uuid):
+    """
+    Deal with database interactions to provide response for get single conversation request.
```
rodriguesk, line 7: Can you add the inputs and outputs to this function description? We want it to be a standard that every function has a description of what the function does as well as the input parameters listed with their description and description of everything returned by the function. Mimic the format used in other newer functions in the app. The older functions might not use the format but we are trying to switch to have all the functions use the documentation format.

user_b/analytics_logging.py
```diff
@@ -0,0 +1,51 @@
...
+from app import db
+from app.errors.errors import DatabaseError
+from app.models import UserBAnalyticsData
+from datetime import datetime, timezone
+from enum import Enum
+
+
+class eventType(Enum):
+    """
+    Event types to be used for analytics logging.
```
rodriguesk, line 12: I think it’s best that analytic event types and names be decided by an analytics team member or in collaboration with analytics team member to minimize need for code changes.

rodriguesk (changes requested):
I like the way you set up the analytics event logging as a single function that can be used for all the user b events. That will be very handy! 

This api endpoint call from what I can tell is just to retrieve details about a conversation so it can be displayed to User A in the expanded conversation card. It’s not for user B clicking the conversation link sent to them by user A (from what I can tell). So I don’t think a user B analytics event is needed? I guess it could be that the link is activated, but there’s no way to know if it has actually been sent to user B yet, so I’m not sure how helpful that would be. But I do think that creating a conversation should be a user A event and that the frontend should probably trigger it through google analytics.

Great having docstring for each new function. Be good to add more to them though: We want to keep docstrings in the standard format that includes description of the function, description of the input parameters, and description of what is returned. Those 3 components are the standard for the Python PEP8 style guide we are following and standard I’ve seen for all programming languages style guides.

Don’t forget to add new endpoints to stoplight documentation. Doing this before coding might be better in the long run so the frontend can comment on it and changes are made before backend coding starts.

We should have a chat for what the backend user b journey event types are and what the names for each event type should be. It’s good that you don’t log event types as numbers because it’ll be a mad house for anyone trying to analyze the analytics data. I think the way you coded it up it replaces the numbers with a description, but I think we need finer resolution for events than just 4 types. I’ll try to make a video describing the possible event types if that would help.

app/conversations/routes.py
y-himanen: No change. I prefer cleaner endpoints.

app/conversations/routes.py
y-himanen: Ok. Got lazy as we write such self-documenting code.

app/conversations/routes.py
y-himanen: This endpoint will be called when user b uses the unique link. It is unclear whether this will ever be part of user a's journey at all at the moment. But this is the structure agreed with frontend at the moment.

I was unaware of the existence of 'the analytics page' and do not want to be responsible for maintaining. The events I coded up in this round are the ones that you and I agreed, and the list is not exhaustive. Analytics do not need to be involved in any naming conventions in the backend. I can, however, change the name of the events going into the tables if you feel that they are not self-explanatory. I personally think they were clear enough, and changing them would be hair-splitting.

app/conversations/utils.py
y-himanen: Yup.

app/conversations/routes.py
y-himanen: Having double-checked, I can see that I misremembered the values for the effect and solutions choices though, so I will update those to UUIDs rather than the IRIs.

rodriguesk (approved):

PR #282, opened 2022-01-26
session uuid validation and analytics tracking removed from get conve…

y-himanen:
…rsation endpoint


cyplas (commented):
This commit is actually also included in PR #283.

PR #283, opened 2022-01-26
Cm 980 982 post user b event

y-himanen:


cyplas (approved):
I've looked over the code and tested it, and it seems fine to me. Just two comments:
1. In clearing the conversations GET docstring, you've also removed mention of conversation uuid validation, even though that is still done.
2. The new user_b_event POST is in line with the stoplight specs, but I've noticed that this would be our first POST route which takes a parameter as a route parameter.

PR #285, opened 2022-02-01
Cm 882 cm 892 get shared impacts

y-himanen:


cyplas (approved):
Looks good. Would just suggest that you add brief code comments for the hard-coded values at lines 174 and 218.

PR #286, opened 2022-02-04
Included all value properties from json

cyplas:


y-himanen (approved):
Good stuff. Would do some renaming on line 51 of app/alignment/utils.py just for the sake of code readability.

PR #287, opened 2022-02-05
added get shared solutions endpoint. edited docstring and added comme…

y-himanen:
…nt for hardcoded values in shared impacts code


cyplas (approved):
Looks and tests good!

PR #298, opened 2022-03-19
added get details endpoints for shared solutions and shared impacts a…

y-himanen:
…nd updateed the user b event endpoint to include more event types

### DIFF
--- a/app/alignment/utils.py
+++ b/app/alignment/utils.py
@@ -299,10 +299,6 @@ def build_shared_solutions_response(alignment_scores_uuid):
         alignment_feed.aligned_solution_7_iri,
     ]
 
-    climate_solutions_iris = [
-        "webprotege.stanford.edu." + iri for iri in climate_solutions_iris
-    ]
-
     climate_solutions = solution_details(G, climate_solutions_iris, nx)
 
     response = {
@@ -324,8 +320,10 @@ def solution_details(G, climate_solutions_iris, nx):
 
     for iri in climate_solutions_iris:
         for node in G.nodes:
-            if G.nodes[node]["iri"] == iri:
-                nx.set_current_node(G.nodes[node])
+            nx.set_current_node(G.nodes[node])
+
+            if nx.get_node_id() == iri:
+
                 solution = {
                     "solutionId": nx.get_node_id(),
                     "sharedScore": 42.00,
@@ -442,8 +440,10 @@ def build_shared_impact_details_response(impact_iri):
     impact = None
 
     for node in G.nodes:
-        if G.nodes[node]["iri"][24:] == impact_iri:
-            nx.set_current_node(G.nodes[node])
+        nx.set_current_node(G.nodes[node])
+
+        if nx.get_node_id() == impact_iri:
+
             if "effect" in G.nodes[node]["all classes"]:
                 impact = {
                     "effectTitle": G.nodes[node]["label"],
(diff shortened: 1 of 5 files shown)


app/alignment/utils.py
```diff
@@ -413,3 +415,93 @@ def log_solution_choice(solution_choice_uuid, shared_solutions):
...
+                "effectSources": nx.get_causal_sources(),
+                "relatedPersonalValues": map_associated_personal_values(
+                    G.nodes[node]["personal_values_10"]
+                ),
+            }
+            break
+
+    if not impact:
+        raise NotInDatabaseError(message="The shared impact IRI cannot be found.")
+    elif len(impact["effectSources"]) == 0:
```
rodriguesk, line 460: Sometimes an impact might not have a source for some reason. Checking to make sure the title exists might be a better way to check that the impact exists?

app/alignment/utils.py
```diff
@@ -413,3 +415,93 @@ def log_solution_choice(solution_choice_uuid, shared_solutions):
...
+
+    for node in G.nodes:
+        if G.nodes[node]["iri"][24:] == solution_iri:
+            nx.set_current_node(G.nodes[node])
+            solution = {
+                "solutionTitle": G.nodes[node]["label"],
+                "imageUrl": nx.get_image_url_or_none(),
+                "longDescription": nx.get_description(),
+                "solutionSources": nx.get_solution_sources(),
+                "solutionType": "mitigation",
```
rodriguesk, line 498: I don’t think this needs to be hard coded. It’s in the class field for the node from what I remember.

app/alignment/utils.py
y-himanen: I meant to ask you if it could be assumed that there are sources. Obviously it can't. I added this check with the thought that if someone mixed up impact and solution IRIs while testing, they would get that message because the function to pull sources for those different types is different, while other functions like for pulling the title are shared.

But note, the first check will see if the IRI exists at all. The second check is only type validation of sorts.

app/alignment/utils.py
y-himanen: I found it hard-coded elsewhere for when we pull the feed. I'll have to look deeper if you say that it should be available in the graph.

rodriguesk (changes requested):
Nice job! There’s just 2 small things to change for pulling information from the networkx graph. 
- the error checking to ensure something a climate impact. Check the node properties or attributes for class field of “effect" or risk.
- ‘Mitigation’ shouldn’t be hard coded. It should exist as a label somewhere in the node properties or attributes.

app/alignment/utils.py
```diff
@@ -413,3 +415,105 @@ def log_solution_choice(solution_choice_uuid, shared_solutions):
...
+    - the sources for the solution
+    - the solution type
+    """
+
+    G = current_app.config["G"].copy()
+    nx = network_x_utils()
+    solution = None
+
+    for node in G.nodes:
+        if G.nodes[node]["iri"][24:] == solution_iri:
```
rodriguesk, line 492: this should use this function instead of hard coding '24:' to get the iri ID https://github.com/ClimateMind/climatemind-backend/blob/0299b7106f42bdb06942a4e7485143c63fb893ce/app/network_x_tools/network_x_utils.py#L23

rodriguesk (approved):
Beautiful!

PR #299, opened 2022-03-19
get alignment summary endpoint added

y-himanen:


app/alignment/utils.py
```diff
@@ -413,3 +413,117 @@ def log_solution_choice(solution_choice_uuid, shared_solutions):
...
+
+    except:
+        raise DatabaseError(
+            message="Something went wrong while retrieving the alignment summary data from the db."
+        )
+
+    return response
+
+
+def get_title_by_iri(iri_list, G, nx):
```
rodriguesk, line 504: I'm thinking this code might be better to live not specifically to this endpoint. I'm thinking it should be added to the network_x_tools folder to reduce code redundancies in the future. [network_x_tools](https://github.com/ClimateMind/climatemind-backend/tree/develop/app/network_x_tools)

app/alignment/utils.py
```diff
@@ -413,3 +413,117 @@ def log_solution_choice(solution_choice_uuid, shared_solutions):
...
+    Returns
+    =========
+    A list of strings.
+    """
+
+    iri_titles = []
+
+    for iri in iri_list:
+        for node in G.nodes:
+            if G.nodes[node]["iri"][24:] == iri:
```
rodriguesk, line 523: this should use this function here I'm thinking instead of hard coding '24:' https://github.com/ClimateMind/climatemind-backend/blob/0299b7106f42bdb06942a4e7485143c63fb893ce/app/network_x_tools/network_x_utils.py#L23

rodriguesk (changes requested):
Looks great! Nice documentation and clean code. There were 2 bits of code regarding networkx utils. Try to add networkx code to the networkx tools utils folder and also leverage code from that folder that's already written. Just trying to keep things as organized as possible. https://github.com/ClimateMind/climatemind-backend/blob/0299b7106f42bdb06942a4e7485143c63fb893ce/app/network_x_tools/network_x_utils.py#L23

rodriguesk (approved):

PR #300, opened 2022-03-19
Cm 1000 fix scoring function

rodriguesk:
Changed 2 things:
- math bug that averaged incorrectly when set 2 of questions was provides
- positivity constant set to 6 instead of 3.5


danmash (approved):
👍

y-himanen (approved):

PR #303, opened 2022-04-03
Cm 931 personalise alignment matching

cyplas:


cyplas:
I wrote short docstrings, as the new functions are more "internal", but I can expand, if desired.
@rodriguesk My docstring for calculate_match is quite generic - maybe you could edit it to explain briefly how this formula came about.

rodriguesk (approved):
Nice work. Especially for your pytest tests! Awesome!

I suggest steering away from standard python lists and dictionaries for these types of numeric scoring processes for our app. Use Numpy. It's at least 30 times faster in most cases. Using Numpy might not be critical for this calculation but for the shared feed calculations it's essential. Numpy does vectorized math, which enables automatic parallelization of these computations. Essentially if you used it here, it would compute all 10 personal value matches at once. I think a separate PR can be used to make this improvement though. It's not essential here, but will be essential for the shared feed computations ticket.

Also, I think it would be good to have the docstring for calculate_match to conform to PEP8 standards since it's the most critical function for this PR. The description is missing information about the the output format (specifically that it's a decimal float value from 0 to 1. And the inputs should be integers.

I don't think it's wise for us to read in the values from the json file more than once when the app gets used. It's probably best we have some constant object persist as some configuration that load when the app starts up that has the value descriptions so it's available at any time when the app is used?  This isn't necessary for this PR, but just a recommendation for the app moving forward. It should be done the next time this get_value_descriptions function is needed.

y-himanen (dismissed):
I agree with everything in the review from @rodriguesk. My only hard request for changes before merging this would be to expand the calculate_match docstring.

In addition, I would be happy if you would consider using more self-explanatory naming in your functions and variables in the spirit of clean code. Longer but clearer names make sense to me, although obviously it is a question of finding a balance :)

cyplas:
Thanks for the feedback! I generally agree with the suggestions too, but I think that maybe some would be better addressed in a codebase cleanup/refactor (e.g., I followed our precedents in how I read from the json file). 

For now I've just addressed Yasmine's hard request, so that you can merge this if you approve without further ado. I/we can address the other points separately (among which, I still suggest, Kameron, that you add something about how the formula came about (whether in the docstring or the google doc), as mentioned earlier).

rodriguesk (approved):

PR #308, opened 2022-04-23
Cm 923 improve recommended solutions

cyplas:
I wrote a basic test for find_alignment_solution_iris, but I am getting a `RuntimeError: Working outside of application context.` even if the test file is absolutely empty. Perhaps I'm not configuring something needed, but interestingly, it seems to work fine if I first merge Daniil's CM-1034-pytest-setup-with-db into my branch. So I haven't committed it and suggest I add it afterwards, not in this pull request.

### DIFF
--- a/app/feed/process_alignment_feed.py
+++ b/app/feed/process_alignment_feed.py
@@ -77,9 +77,9 @@ def find_alignment_solution_iris(
     """Choose and order solutions for the alignment.
 
     Using the (mitigation) solutions from the ontology, put the conversation solution first,
-    followed by a random ordering of POPULAR_SOLUTION_COUNT solutions from POPULAR_SOLUTION_NAMES
-    and UNPOPULAR_SOLUTION_COUNT other solutions. This function takes no arguments, since solutions
-    are (currently) independent of users' personal values etc.
+    followed by a random ordering of popular and unpopular solutions. This function takes no
+    user-specific arguments, since solutions are (currently) independent of users' personal values
+    etc.
 
     Parameters
     ==============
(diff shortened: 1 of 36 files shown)


app/feed/process_alignment_feed.py
```diff
@@ -1,38 +1,100 @@
...
+
+    Returns
+    ==========
+    List of strings: an ordered list of solution iris for an alignment feed
+    """
+    solution_map = {"talk": None, "popular": set(), "unpopular": set()}
+    solution_nodes = get_solution_nodes()
+    for node in solution_nodes:
+        name = node["label"]
+        iri = node["iri"][len("webprotege.stanford.edu.") :]
```
rodriguesk, line 82: There’s a helper function already somewhere in the codebase to do this line. You should just use it here since it is possible that the iri base could change in the future. It will minimize the places need to change webprotege.stanford.edu

rodriguesk (approved):
Nice work!

I recommend changing the 1 line comment I mentioned (using a helper function already written for obtaining the formatted iri string from a node in the ontology).

Also, generally speaking I recommend even constants to be passed into functions as input arguments so that it’s easier to understand where things are coming from and it helps reduce dependency on memorizing constants that might have a global scope for a module. I guess it’s fine here, but might be better if loaded these global constants for the config file if you truly don’t want to have these constants be input parameters for the function getting solutions for User B.

cyplas:
> I recommend changing the 1 line comment I mentioned (using a helper function already written for obtaining the formatted iri string from a node in the ontology).
> 
> Also, generally speaking I recommend even constants to be passed into functions as input arguments so that it’s easier to understand where things are coming from and it helps reduce dependency on memorizing constants that might have a global scope for a module. I guess it’s fine here, but might be better if loaded these global constants from the config file (similar to how accessing the ontology) if you truly don’t want to have these constants be input parameters for the function getting solutions for User B.

Ok, I've tweaked and pushed.
I haven't put the constants in the global constants file, as currently that file doesn't include such specific settings, so perhaps it's better if that's part of our general backend refactoring/cleanup plans.

rodriguesk (approved):

PR #309, opened 2022-04-24
CM-998 sessions module refactoring and unit & integration tests added.

danmash:
- small refactoring `post_session` endpoint.
- unit & integration tests

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1,5 +1,6 @@
 import os
 from flask import current_app
+from sqlalchemy.orm import relationship
 from sqlalchemy.orm.session import close_all_sessions
 from app.extensions import db, jwt
 from werkzeug.security import generate_password_hash, check_password_hash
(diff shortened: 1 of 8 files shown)


app/models.py
```diff
@@ -16,6 +17,7 @@ class Sessions(db.Model):
     __tablename__ = "sessions"
     ip_address = db.Column(db.String(255), primary_key=False)
     user_uuid = db.Column(UNIQUEIDENTIFIER, db.ForeignKey("users.user_uuid"))
+    user = relationship("Users", foreign_keys=[user_uuid])
```
danmash, line 20: To use subfactory in SeessionFactory

app/session/routes.py
```diff
@@ -24,8 +23,8 @@ def post_session():
     if current_user:
         user_uuid = current_user.user_uuid
 
-    store_session(session_uuid, session_created_timestamp, user_uuid)
-    process_ip_address(request, session_uuid)
+    ip_address = get_ip_address(request)
+    store_session(session_uuid, session_created_timestamp, user_uuid, ip_address)
```
danmash, line 27: instead of 
- create
- get IP 
- save again
we can use just: 
- get IP
- create

rodriguesk (approved):

PR #313, opened 2022-04-26
CM-1027 questions JSON schema added for validation, JSON file read mo…

danmash:
CM-1027
- questions JSON schema added for validation,
- JSON file read moved to the utils.py. 
- unit & integration tests added


rodriguesk (approved):
I like how you moved the loading of questions to the config, because that’s essential for the app to work. The name change to the old function makes sense also because the purpose of the function is to get the data from the questions file.

Nice job!

rodriguesk (approved):

PR #316, opened 2022-04-26
CM-1017 auth/utils moved to common/ with UTs

danmash:
moved because these functions are not related to auth


app/common/local.py
```diff
@@ -0,0 +1,20 @@
...
+    A helper function to check whether a request is being made from a local connection or on the
+    live website.
+    """
+    local = False
+
+    if os.environ.get("IS_LOCAL"):
+        local = (
+            request.remote_addr == "127.0.0.1"
+            or os.environ.get("VPN")
+            or request.host == "localhost"
```
danmash, line 17: @y-himanen I need this like for pytest. Do I understand correctly that on prod it will be different? Just to make sure

app/common/uuid.py
```diff
@@ -0,0 +1,73 @@
+import typing
+import uuid
+from enum import Enum
+
+from app.errors.errors import InvalidUsageError, NotInDatabaseError
+from app.models import Sessions, Scores, Users, Conversations, AlignmentScores
+
+
+class uuidType(Enum):
```
rodriguesk, line 9: This has to do with the conversation dashboard. I don’t know if I’d consider it ‘common’ to be honest!

rodriguesk (changes requested):
It is a major change to move all of the utility functions to a common place. But I guess there are so many instances in the app where we check authorization, so it might make sense to move it there. The conversation status thing isn’t going to be common. It’ll be related to a feature we haven’t fully built up yet that have to do with allowing user A to track conversations they have with different people.

app/common/local.py
danmash: @rodriguesk This is clear. The question is `request.host == "localhost"` will be false on prod or not? I suppose it will be false, but I want just to double-check.

danmash:
> The conversation status thing isn’t going to be common. So my only requested changes are to move that to somewhere better than common or to leave that where it currently is.

@rodriguesk could you please clarify? I didn't move the `app.conversations.routes.ConversationStatus` is still on the same place.

rodriguesk:
@danmash i was just referring to the status check for enums type. That was written to be for error checking, but it’s just used as part of user b journey set of features. You can move it to common but I can see it easily becoming hard to find. This might be hinting that some better organization system might be needed for the common user b functions.

danmash:
@rodriguesk sorry, I didn't get it. Could you please comment on the exact line https://github.com/ClimateMind/climatemind-backend/pull/316/files 
or push your idea directly to this branch if it's not presented in this PR

rodriguesk:
It’s for the comment you mark resolved. It was that line.

app/common/local.py
rodriguesk: yes, it should be.

rodriguesk:
I'm referring to the following code, that it might get used in multiple places but it is actually just for conversation features, it's not across the whole app. It should only be used for when the user b takes the quiz and shares, basically. I think there's many functions that have been written to wrong folders for the conversation features. I'm saying that it might be better to make a separate folder or subfolder or even a list someplace else that's for code to relook at later for refactoring later after you've gone through the whole app. I'm pretty sure this one belongs in some conversations features folder or a folder not yet made that's conversation_dashboard. 

class uuidType(Enum):
Show resolved
    """
    UUID types used to print for error responses
    """

    SESSION = 1
    QUIZ = 2
    USER = 3
    CONVERSATION = 4
    ALIGNMENT_SCORES = 5

rodriguesk:
Nevermind, I think I got confused with something else. This is all good. Approved.

rodriguesk (approved):

PR #317, opened 2022-04-28
CM-1038 alignment feed bug and logic fix

rodriguesk:
The bug was found and fixed. It had to do with checking for a value in the database that had not yet been written to the db. In addition, I fixed some missing logic that didn’t get caught during PR of CM-921. The top 3 shared topics are supposed to be ordered via version 1 scoring using user b’s scores, not topics with the shared personal value.

Lastly, the results are returned as lists instead of dictionaries since only iri are needed. This is minor, but necessary because if there are less than 3 topics possible, 1 topic needs to be repeated and if dictionaries are used then the topic can’t be repeated because dictionaries can’t have duplicate keys.

### DIFF
--- a/app/feed/process_alignment_feed.py
+++ b/app/feed/process_alignment_feed.py
@@ -36,12 +36,15 @@ def create_alignment_feed(
     quiz_uuid (UUID) - user b quiz uuid to compare scores with user a scores
     alignment_feed_uuid (UUID) - uuid created when post alignment endpoint is used
     """
-    
 
-    #TO DO make a contant variable so 3 isn't hard coded as the number of effects to show to user B.
-    aligned_effects_sorted_by_shared_values = get_aligned_effects(alignment_scores_uuid, 3)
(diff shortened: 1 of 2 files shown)


app/feed/process_alignment_feed.py
```diff
@@ -37,15 +37,14 @@ def create_alignment_feed(
     alignment_feed_uuid (UUID) - uuid created when post alignment endpoint is used
     """
 
-    aligned_effects = list(get_aligned_effects(alignment_scores_uuid).keys())
-
-    sorted_aligned_effects = list(
-        sort_aligned_effects_by_user_b_values(aligned_effects, conversation_uuid).keys()
+    # TO DO make a contant variable so 3 isn't hard coded as the number of effects to show to user B.
```
danmash, line 40: Make sense, I think we could move it to `DevelopmentConfig` right now.

danmash (approved):

PR #318, opened 2022-05-01
CM-923: added test

cyplas:
Added basic pytest test for main function from CM-923.

### DIFF
--- a/app/feed/process_alignment_feed.py
+++ b/app/feed/process_alignment_feed.py
@@ -52,15 +52,11 @@ def create_alignment_feed(
         alignment_feed.aligned_effect_1_iri = sorted_aligned_effects[0]
         alignment_feed.aligned_effect_2_iri = sorted_aligned_effects[1]
         alignment_feed.aligned_effect_3_iri = sorted_aligned_effects[2]
+        temporary_solutions_iris = get_default_solutions_iris()
         assign_alignment_iris(
             alignment_feed,
             "solution",
-            find_alignment_solution_iris(
-                CONVERSATION_SOLUTION_NAME,
-                POPULAR_SOLUTION_NAMES,
-                POPULAR_SOLUTION_COUNT,
-                UNPOPULAR_SOLUTION_COUNT,
-            ),
+            temporary_solutions_iris,
         )
         db.session.add(alignment_feed)
         db.session.commit
@@ -149,22 +145,13 @@ def assign_alignment_iris(alignment_feed, field_type, iris):
         setattr(alignment_feed, "aligned_{}_{}_iri".format(field_type, index), iri)
 
 
-def find_alignment_solution_iris(
-    conversation_name, popular_names, popular_count, unpopular_count
-):
+def get_default_solutions_iris() -> list:
     """Choose and order solutions for the alignment.
 
     Using the (mitigation) solutions from the ontology, put the conversation solution first,
     followed by a random ordering of popular and unpopular solutions. This function takes no
     user-specific arguments, since solutions are (currently) independent of users' personal values
     etc.
-
-    Parameters
-    ==============
-    conversation_name (str) - label of the conversation solution
-    popular_names (set of str) - labels of the popular solutions
-    popular_count (int) - the number of popular solutions to include
-    unpopular_count (int) - the number of unpopular solutions to include
 
     Returns
     ==========
@@ -174,20 +161,23 @@ def find_alignment_solution_iris(
     solution_map = {"conversation": None, "popular": set(), "unpopular": set()}
     solution_nodes = get_solution_nodes()
     nx = network_x_utils()
+
     for node in solution_nodes:
         name = node["label"]
         nx.set_current_node(node)
         iri = nx.get_node_id()
-        if name == conversation_name:
+        if name == CONVERSATION_SOLUTION_NAME:
             solution_map["conversation"] = iri
-        elif name in popular_names:
+        elif name in POPULAR_SOLUTION_NAMES:
             solution_map["popular"].add(iri)
         else:
             solution_map["unpopular"].add(iri)
-    sample_solutions = sample(solution_map["popular"], popular_count) + sample(
-        solution_map["unpopular"], unpopular_count
+
+    sample_solutions = sample(solution_map["popular"], POPULAR_SOLUTION_COUNT) + sample(
+        solution_map["unpopular"], UNPOPULAR_SOLUTION_COUNT
     )
     shuffle(sample_solutions)
+
     return [solution_map["conversation"]] + sample_solutions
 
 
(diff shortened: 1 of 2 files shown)


rodriguesk (approved):

app/feed/process_alignment_feed.py
```diff
@@ -55,12 +55,7 @@ def create_alignment_feed(
         assign_alignment_iris(
             alignment_feed,
             "solution",
-            find_alignment_solution_iris(
-                CONVERSATION_SOLUTION_NAME,
-                POPULAR_SOLUTION_NAMES,
-                POPULAR_SOLUTION_COUNT,
-                UNPOPULAR_SOLUTION_COUNT,
-            ),
+            find_alignment_solution_iris(),
```
danmash, line 58: I guess if we pass there constants only we could omit arguments at all

app/feed/process_alignment_feed.py
```diff
@@ -174,20 +160,23 @@ def find_alignment_solution_iris(
     solution_map = {"conversation": None, "popular": set(), "unpopular": set()}
     solution_nodes = get_solution_nodes()
     nx = network_x_utils()
```
danmash, line 162: A general recommendation is splitting logical blocks with a newline.

app/feed/tests/test_process_alignment_feed.py
```diff
@@ -0,0 +1,53 @@
...
+        iri_name_map[nx.get_node_id()] = node["label"]
+
+    result_solution_iris = find_alignment_solution_iris()
+    conversation_solution_count = 1
+    total_solutions_count = (
+        POPULAR_SOLUTION_COUNT + UNPOPULAR_SOLUTION_COUNT + conversation_solution_count
+    )
+    assert (
+        len(result_solution_iris) == total_solutions_count
+    ), f"Total solutions cound should be {total_solutions_count}"
```
danmash, line 29: Don't move all asserts to the bottom, try to keep them near variables you use per assert

app/feed/tests/test_process_alignment_feed.py
```diff
@@ -0,0 +1,53 @@
...
+
+    first_iri_in_result = result_solution_iris[0]
+    assert (
+        iri_name_map[first_iri_in_result] == CONVERSATION_SOLUTION_NAME
+    ), "First iri in result should be CONVERSATION iri"
```
danmash, line 34: Use "comments" after assert to increase readability of the test and highlight the purpose of the test

danmash:
@cyplas @rodriguesk I've updated the test and added some recommendations ( + refactor a little). does it make sense?

danmash (approved):

app/feed/process_alignment_feed.py
rodriguesk: I talked about this before with Cyprian. The problem is the constants aren’t set in the app config so it can be hard to know they exist and hard to know where to find them to change them. So we kept them as inputs just to make it clearer that constants were getting used as arguments. Generally, I think it’s bad style to use scoping to keep track of access to constants. It’s better to have constants defined close to where they get used or to have them be app-wide constants. If both you and Cyprian prefer no arguments then we can just go with consensus instead of my concerns about style and scoping.

danmash: well, since it's not personalized at all and will be improved in the future (I hope so, otherwise it doesn't make sense) I would keep it as is for now.
also, I suggest to rename it to avoid confusion. 

I thought this part of the project is crucial to have to be able to provide meaningful conversations. when we are planning to adjust this?

danmash (approved):

rodriguesk:
The effect topics are personalized, but the solution topics aren’t at the
moment.

Once we have enough data on which topics are most engaged with by which
types of people given their personal values profiles, then we can start
doing personalized recommendations. the data we collect with this could
will enable us to personalize in the future because there’s a bit of
randomness to what we are recommending, while we still are using the most
popular topics from version 1 (so it’s not totally random).

danmash:
@rodriguesk 

>  Generally, I think it’s bad style to use scoping to keep track of access to constants.

sure, I think if we will use these constants somewhere else we need to move it to config. But it's ok at the moment since there's only one usage. I going to merge it as is right now.

app/feed/process_alignment_feed.py
cyplas: > Generally, I think it’s bad style to use scoping to keep track of access to constants. It’s better to have constants defined close to where they get used or to have them be app-wide constants.

I think what happened here is that I assumed you weren't initially talking about app-wide constants, as I didn't find similar precedents in the codebase, and I assumed you weren't talking about constants declared inside functions, as then I don't think of them as "constants", which left me with this awkward middle ground. :) But indeed, we can improve this at another time.

rodriguesk:
Yeah, it’s minor, nothing to waste a lot of time over. The name changes
Daniil made helped reduce my worry about hidden constants and hidden usage
of them too. Which is good.

danmash:
> The effect topics are personalized, but the solution topics aren’t at the
> moment.
> 
> Once we have enough data on which topics are most engaged with by which
> types of people given their personal values profiles, then we can start
> doing personalized recommendations.

 @rodriguesk maybe it's a good idea to synchronize proposed actions with https://www.un.org/actnow ?

rodriguesk:
We have the actions pulled  currently from Project Drawdown (I talked to them and got their permission to use their text from their website). It’s a good idea you have though, so I added it to a resource file I have for all the apps and websites that help offer more actions.

danmash:
I'm sure we have all these actions already. I'm talking about  the  set of
actions to show to the user instead of randomized like how

rodriguesk:
Are you thinking you’d like to see the same 10 things recommended as individual actions to every user? Like this graphic? https://www.un.org/sites/un2.un.org/files/actnow_action_guide_2021.pdf

There’s climate solutions that are individual actions and there’s ones that are system/societal changes/policy. We aim to show both types to users, as there’s different arguments for both individual action and societal change. Currently we show 7 actions (one or which is always to have climate conversations) and by having solutions change for each conversation with user Bs we are helping incentivize user A more conversations with people to get exposed to more solutions. In addition to showing everyone talking about climate change, we also show 3 random solutions from individual actions such as :
- composting
- eating lower down the food chain
- talking about climate change
- etc (these could pull from a pull of 10 individual actions like you said if that’s what you’re suggesting?)

Then we also show 3 random system/societal changes/policy that have been calculated by Project Drawdown to be the most impactful. These are things people can’t do individually as one person. So it requires exposure to lots of people to build support (system change).

What are you suggesting exactly? To shrink the pools of individual actions and societal actions? If we do that it will make the solutions offered for every conversation to be the same, which would get pretty boring I think for alarmed people talking to others. It would also reduce learning opportunities for alarmed people when they talk to user B. 

That said, we could definitely expand options and weight things if you’re thinking of that.

The solutions shown are completely random though. They are prioritized in a data-driven way.

PR #323, opened 2022-05-13
CM-1015

danmash:
- get_alignment integration and UT added
- new factories added

### DIFF
--- a/app/alignment/routes.py
+++ b/app/alignment/routes.py
@@ -91,6 +91,7 @@ def get_alignment(alignment_scores_uuid):
     - user b's name
     """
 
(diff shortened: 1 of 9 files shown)


rodriguesk (approved):
I'm not sure what the SQLalchemy 'relationship()' do in app/models.py
Let me know and make sure they don't cause any issues. I'm not sure if we didn't have them for a reason or if it just wasn't use out of not knowing they should be used or are useful.

danmash:
I use `relationship` for an automatic sub-factories creation with `factoryboy`. The first such `ralationship` was added by me a month ago to a `Session` object. No issues so far.

PR #325, opened 2022-05-15
Cm 1044 personal value representation

danmash:
There are two endpoints with topMatchValue and I don't know which one is used on that page so I've changed both.

/alignment/<alignment_scores_uuid>
/alignment/<alignment_scores_uuid>/summary

Thus we need to test the usage of these two endpoints. 

The problem is that we are storing the personal value `key` in the database in `AlignmentScore` object, (which is not good btw). This key should be transformed into some representation. I did this in `PersonalVaue` since it's the single place for all `PersonalValue` stuff. 

@cyplas I've tried to change the response example of these two endpoints in stoplight but it affected a lot of fields, I don't know why. could you have a look?
Screenshot 2022-05-15 at 22 42 01

### DIFF
--- a/app/static/Climate-Mind_bundled.yml
+++ b/app/static/Climate-Mind_bundled.yml
@@ -99,10 +99,7 @@ paths:
                     conversations:
                       - null
       operationId: get-conversations
-      description: >-
-        Registered users can see a list of their conversations. Returns a list
-        of the user's conversations with the relevant names, UUIDs, conversation
-        status and creation date. A JWT token is required.
+      description: 'Registered users can see a list of their conversations. Returns a list of the user''s conversations with the relevant names, UUIDs, conversation status and creation date. A JWT token is required.'
       parameters:
         - schema:
             type: string
@@ -137,9 +134,7 @@ paths:
                     message: conversation status updated
                     conversationId: '{conversation_uuid}'
       operationId: put-conversation
-      description: >-
-        Change the status of the conversation in the conversations table for
-        User A according to the actions taken by User B.
+      description: Change the status of the conversation in the conversations table for User A according to the actions taken by User B.
       parameters:
         - schema:
             type: string
@@ -283,9 +278,7 @@ paths:
                   value:
                     alignmentScoresId: d86fcf26-7834-4bce-bfb7-f88170da9ace
       operationId: post-alignment
-      description: >-
-        To create aligned scores between user A and user B after user B has
-        completed their quiz.
+      description: To create aligned scores between user A and user B after user B has completed their quiz.
       parameters:
         - schema:
             type: string
@@ -421,7 +414,7 @@ paths:
                   value:
                     overallSimilarityScore: 70
                     topMatchPercent: 90
-                    topMatchValue: benevolence
+                    topMatchValue: Benevolence
                     valueAlignment:
                       - description: lots of text
                         id: benevolence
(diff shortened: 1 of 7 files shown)


app/static/Climate-Mind_bundled.yml
```diff
@@ -421,7 +414,7 @@ paths:
                   value:
                     overallSimilarityScore: 70
                     topMatchPercent: 90
-                    topMatchValue: benevolence
+                    topMatchValue: Benevolence
```
danmash, line 417: This is my change, I don't know why there are so much stuff committed with it

app/static/Climate-Mind_bundled.yml
```diff
@@ -1360,23 +1237,20 @@ paths:
                 Successful GET Alignment Summary Response Body:
                   value:
                     userAName: Stevie
-                    topMatchValue: benevolence
+                    topMatchValue: Self Direction
```
danmash, line 1240: And this is also my change

cyplas:
> @cyplas I've tried to change the response example of these two endpoints in stoplight but it affected a lot of fields, I don't know why. could you have a look?

@danmash I'm not sure what you mean. The only difference I see here between the develop and CM-1044 branches is in the topMatchValue, in line with the diff you show below. Maybe your browser is doing something funny?

danmash:
@cyplas please have a look at the diff I see, it's enormous. 
[Cm 1044 personal value representation by danmash · Pull Request #325 · ClimateMind_climatemind-backend.pdf](https://github.com/ClimateMind/climatemind-backend/files/8703448/Cm.1044.personal.value.representation.by.danmash.Pull.Request.325.ClimateMind_climatemind-backend.pdf)

cyplas:
> @cyplas please have a look at the diff I see, it's enormous.

@danmash Indeed, the diff is big, but you mentioned changed fields, and on quick perusal, I don't see that (neither in the diff nor on the webpage). It looks to me like Stoplight has changed how it's formatting the yml file (whitespace, quotes, etc), and it has done so also in at least one of my stoplight commits. In short, I don't know what's up with this, but I suggest if it doesn't keep happening "randomly", we don't worry about it. Unless you've noticed it changing things that aren't just harmless reformatting ...

danmash:
> > @cyplas please have a look at the diff I see, it's enormous.
> 
> 
> 
> @danmash Indeed, the diff is big, but you mentioned changed fields, and on quick perusal, I don't see that (neither in the diff nor on the webpage). It looks to me like Stoplight has changed how it's formatting the yml file (whitespace, quotes, etc), and it has done so also in at least one of my stoplight commits. In short, I don't know what's up with this, but I suggest if it doesn't keep happening "randomly", we don't worry about it. Unless you've noticed it changing things that aren't just harmless reformatting ... 

If so, could you please approve this pr?

danmash:
@rodriguesk I think this is ready to merge

rodriguesk (approved):

app/static/openapi.yaml
```diff
@@ -1,15 +1,13 @@
...
+  description: The API for sending and recieving information to and from the ClimateMind ontology.
+  version: '0.1'
 servers:
-  - url: http://127.0.0.1:5000
-  - url: http://localhost:5000
+  - url: 'http://127.0.0.1:5000'
+  - url: 'http://localhost:5000'
 paths:
-  /scores:
+  /alignment:
```
rodriguesk, line 10: This should remain /scores. It should not be /alignment. Right @danmash ?

danmash: you are right

rodriguesk:
@danmash it still looks like /scores is changed to /alignment ?

rodriguesk (changes requested):
/alignment needs to be changed back to say /scores in app/static/openapi.yaml as we agreed in our github comments conversation.

danmash:
@rodriguesk I can't do this in stoplight

Screenshot 2022-05-18 at 23 46 31

rodriguesk:
@danmash endpoint documentation changes must be separate branches than coding changes. Would it help to create a separate branch from this documentation change?

danmash:
I've just updated the file manually, please merge the branch

rodriguesk (approved):

PR #327, opened 2022-05-21
Cm 1042 user a see alignment backend

cyplas:
CM-1042:
 - modified GET /conversation/:id endpoint, so that it conditionally includes alignment id


rodriguesk (commented):
@cyplas do you think it might be better to always have the alignmentId field and just have it be empty/null for when consent=False? I wonder if that would make it easier for frontend to manage?

app/static/Climate-Mind_bundled.yml
```diff
@@ -170,12 +170,28 @@ paths:
...
+                  example-1:
+                    conversationId: cc63e48f-d066-44e7-851c-c28af17ab3fb
+                    userA:
+                      name: Bill
+                      id: ba5df442-7261-4fc1-bff0-5dfd84035d56
+                      sessionId: 671e4949-a3e4-4844-b9d2-cd843f48f357
+                    userB:
+                      name: Bob
+                    conversationStatus: 2
+                    consent: true
```
rodriguesk, line 183: Is this meant to say consent=false? (Instead of true)

app/static/Climate-Mind_bundled.yml
```diff
@@ -235,7 +250,7 @@ paths:
                     userB:
                       name: Bob
                     conversationStatus: 2
-                    consent: true
+                    consent: false
```
rodriguesk, line 253: Is this one meant to say consent= true?

rodriguesk (commented):
I guess we just need to confirm with frontend before merging if they prefer either:
1. field key and value to always exist from an endpoint response and when consent = false to have alignmentId = Null
2. Field key and value pair only to exist conditional on consent=true. If consent=false then the alignmentId field isn’t even included in the endpoint response

@joaorpereira which do you prefer or think is best standard for the frontend to receive?

danmash (dismissed):

joaorpereira:
> I guess we just need to confirm with frontend before merging if they prefer either:
> 
> 1. field key and value to always exist from an endpoint response and when consent = false to have alignmentId = Null
> 2. Field key and value pair only to exist conditional on consent=true. If consent=false then the alignmentId field isn’t even included in the endpoint response
> 
> @joaorpereira which do you prefer or think is best standard for the frontend to receive?

I prefer 1

rodriguesk (changes requested):
@cyplas can you change this endpoint so that it works the following way:
field key and value to always exist from an endpoint response and when consent = false to have alignmentId = Null
By null I think we just mean blank.

cyplas:
> @cyplas can you change this endpoint so that it works the following way: field key and value to always exist from an endpoint response and when consent = false to have alignmentId = Null By null I think we just mean blank.

Ok, will do. And I agree that this is probably better.

app/static/Climate-Mind_bundled.yml
cyplas: Hmm, this is under x-examples:example-1, which I don't think I created and I don't see in the interface. Unless I'm missing something, I suggest I add this to my low-priority todos.

app/static/Climate-Mind_bundled.yml
cyplas: No, this is under the "Successful GET Conversation Response Body with consent=false" example.

cyplas:
> field key and value to always exist from an endpoint response and when consent = false to have alignmentId = Null
By null I think we just mean blank.

Added this now. I interpreted null here as the (as far as I understand) standard json unquoted null, equivalent to python None. Also included now are those automatic stoplight js/json files, as we agreed.

PR #329, opened 2022-05-24
Cm 939 backend conversation edit endpoint

danmash:
I've created an endpoint https://climatemind.stoplight.io/docs/climatemind-backend/branches/develop/a79c424fa467d-single-conversation-get (docs will be updated tomorrow) 

I think it's essential to bring a [serialization](https://www.django-rest-framework.org/api-guide/serializers/) concept to the project. In a nutshell, it should do almost all the work from our `utils.py` files including:
- processing user input and validation (e.g. validate_uuid)
- preparing response based on DB objects. 

I've noticed we already have a great dependency [marshmallow](https://github.com/marshmallow-code/marshmallow) in the [project](https://github.com/ClimateMind/climatemind-backend/blob/develop/requirements.txt#L50).  but we don't use it for some reason. 
To make it easier to use with flask I've added a couple of more dependencies 
- https://marshmallow-sqlalchemy.readthedocs.io/en/latest/
- https://flask-marshmallow.readthedocs.io/en/latest/
and made it work with PyTest

### DIFF
--- a/app/conversations/schemas.py
+++ b/app/conversations/schemas.py
@@ -1,0 +1,17 @@
+from marshmallow import fields, validate
+
+from app.common.schemas import CamelCaseSchema, ma
+from app.conversations.enums import ConversationStatus
+from app.models import Conversations
+
+
+class ConversationEditSchema(CamelCaseSchema, ma.SQLAlchemySchema):
+    class Meta:
+        model = Conversations
+        load_instance = True
+
+    conversation_id = ma.auto_field("conversation_uuid")
+    receiver_name = ma.auto_field()
+    conversation_status = fields.Integer(validate=validate.OneOf(
+        [s.value for s in ConversationStatus]
+    ))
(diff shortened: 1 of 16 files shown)


app/conversations/schemas.py
```diff
@@ -0,0 +1,17 @@
...
+from app.conversations.enums import ConversationStatus
+from app.models import Conversations
+
+
+class ConversationEditSchema(CamelCaseSchema, ma.SQLAlchemySchema):
+    class Meta:
+        model = Conversations
+        load_instance = True
+
+    conversation_id = ma.auto_field("conversation_uuid")
```
danmash, line 13: altering DB field with a new name

app/conversations/schemas.py
```diff
@@ -0,0 +1,17 @@
...
+    receiver_name = ma.auto_field()
```
danmash, line 14: direct map from the DB field, we could map all the fields using `ma.SQLAlchemyAutoSchema`

app/conversations/schemas.py
```diff
@@ -0,0 +1,17 @@
...
+    conversation_status = fields.Integer(
+        validate=validate.OneOf([s.value for s in ConversationStatus])
+    )
```
danmash, line 17: classic `marshmallow` fields with custom validation

rodriguesk (approved):

rodriguesk (approved):

PR #330, opened 2022-05-27
Cm 1045 db migrations convo state and ratings

rodriguesk:


migrations_local/versions/11bef167c5c2_add_all_current_tables.py
```diff
@@ -174,113 +212,220 @@ def upgrade():
...
+        #     ["scores.quiz_uuid"],
+        # ),
+        # sa.ForeignKeyConstraint(
+        #     ["solution_choice_uuid"],
+        #     ["solution_choice.solution_choice_uuid"],
+        # ),
         sa.PrimaryKeyConstraint("conversation_uuid"),
     )
+
+    op.create_foreign_key(
```
rodriguesk, line 242: I moved this part to be here instead of down below like it was in the previous migration script.

danmash (approved):
👍

PR #332, opened 2022-05-30
CM-683 - CM-683-logged-in-user-can-change-their-password

danmash:
- [ ] docs 

CM-683 
- update-password endpoint was added with integration test
- emails app renamed to account due to work with user account


rodriguesk (changes requested):
Missing minimum password requirement I believe. Please use these: https://github.com/ClimateMind/climatemind-backend/blob/c347d32ff13951157d728a9fe759d286cf6bd3e7/app/auth/routes.py#L278

rodriguesk (approved):
awesome job! Just needs documentation now.

PR #339, opened 2022-06-13
CM-522 these migrations is not needed anymore, job is done long time ago

danmash:


rodriguesk:
I agree this code isn’t an active part of this repo, but I think it was getting kept just in case we need to do it again? Something about how the way LRF table is stored not being perfect and that if we need to update it then we have to drop the whole table and re add the whole table to add another column? I can’t remember exactly.

rodriguesk (changes requested):
Please move these code to an archive folder instead for now. See my comment from yesterday for more details.

danmash:
@rodriguesk I think you misunderstood these migrations scripts. These scripts are particular and not related to LRF at all.
- first one is about replacing `id` with `uuid` fields 
- the second is about moving postal code to different DB table.
These scripts are not suitable for any other task and couldn't be reused in any way.
It would be easier to write a new script if the similar task will appear again.
Anyway, the best `archive` table is `.git` 😉

PR #340, opened 2022-06-20
Cm 1049 conversations get put rebased

danmash:
I've created a new branch based on @cyplas CM-1049-conversation-get-and-put
and changed conversations PUT endpoint

### DIFF
--- a/app/conversations/enums.py
+++ b/app/conversations/enums.py
@@ -1,4 +1,4 @@
-from enum import IntEnum
+from enum import IntEnum, Enum
 
 
 class ConversationStatus(IntEnum):
@@ -6,6 +6,7 @@ class ConversationStatus(IntEnum):
     Conversation status is used to identify where a user is
     in their journey to communicate with other users they invite.
 
+    # FIXME: deprecated
     This enum should not be modified unless the frontend is involved in the change.
     """
 
@@ -17,12 +18,27 @@ class ConversationStatus(IntEnum):
 
 class ConversationState(IntEnum):
     """
-    State which identifies how far along user A is in seeing the conversation results with user B.
+    State which identifies how far along user A is in seeing
+    the conversation results with user B.
     """
 
-    InvitedUserB = 0
-    UserBDone = 1
+    UserBInvited = 0
+    UserBConsented = 1
     AlignButtonClicked = 2
-    TalkButtonClicked = 3
-    DoneButtonClicked = 4
-    ConversationRated = 5
+    TopicsButtonClicked = 3
+    TalkedButtonClicked = 4
+    RatingDone = 5
+
+
+class ConversationButtonMapState(IntEnum):
+    aligned = ConversationState.AlignButtonClicked
+    topics = ConversationState.TopicsButtonClicked
+    talked = ConversationState.TalkedButtonClicked
+
+
+class ConversationUserARating(IntEnum):
+    AWFUL = 1
+    BAD = 2
+    AVERAGE = 3
+    GOOD = 4
+    EXCELLENT = 5
--- a/app/conversations/enums.py
+++ b/app/conversations/enums.py
@@ -13,3 +13,16 @@ class ConversationStatus(IntEnum):
     Visited = 1
     QuizCompleted = 2
     ConversationCompleted = 3
+
+
+class ConversationState(IntEnum):
+    """
+    State which identifies how far along user A is in seeing the conversation results with user B.
+    """
+
+    InvitedUserB = 0
+    UserBDone = 1
+    AlignButtonClicked = 2
+    TalkButtonClicked = 3
+    DoneButtonClicked = 4
+    ConversationRated = 5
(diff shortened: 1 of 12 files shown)


app/conversations/enums.py
```diff
@@ -1,15 +1,44 @@
...
+class ConversationButtonMapState(IntEnum):
+    aligned = ConversationState.AlignButtonClicked
+    topics = ConversationState.TopicsButtonClicked
+    talked = ConversationState.TalkedButtonClicked
+
+
+class ConversationUserARating(IntEnum):
+    AWFUL = 1
+    BAD = 2
+    AVERAGE = 2
```
rodriguesk, line 42: These numbers seem wrong. Please double check these.

app/conversations/enums.py
```diff
@@ -1,15 +1,44 @@
...
+    GOOD = 3
```
rodriguesk, line 43: I think this should be 4 ?

app/static/Climate-Mind_bundled.yml
```diff
@@ -75,6 +46,36 @@ paths:
                         createdByUserId: 39229453-E000-4801-91CA-D1F85E750508
                         createdDateTime: 'Wed, 05 Jan 2022 06:44:33 GMT'
                         invitedUserName: Jim 4
+                properties:
```
rodriguesk, line 49: The examples used in the documentation should match these changes shown here too. I believe the examples shown here don’t match with your changes. Can you make them match?

rodriguesk:
@cyplas can you please review the docs for this PR? I don’t feel confident enough in me judging this. We need to make sure sure the logic is perfect for what the frontend needs for the GET output and the PUT input requirements. That’s what matters most.

rodriguesk (changes requested):
See my inline comments on the github summary screen.
- Documentation needs improving from what I can tell
- errors with enums
- requirements logic needs to be approved by Cyprian

app/conversations/enums.py
danmash: Nice catch!

app/conversations/enums.py
danmash: Nice catch!

app/static/Climate-Mind_bundled.yml
danmash: @cyplas I know you've asked to change docs for the endpoint but I worked on PUT endpoint instead and didn't touch the GET endpoint to which this example belongs to. 
Since you will be around next week, could you please change this yourself? 
Thanks!

app/conversations/enums.py
```diff
@@ -1,15 +1,44 @@
...
+class ConversationButtonMapState(IntEnum):
+    aligned = ConversationState.AlignButtonClicked
+    topics = ConversationState.TopicsButtonClicked
+    talked = ConversationState.TalkedButtonClicked
+
+
+class ConversationUserARating(IntEnum):
+    AWFUL = 1
+    BAD = 2
+    AVERAGE = 3
```
rodriguesk, line 42: this should probably say OKAY instead of AVERAGE. We hope the mean conversations are GOOD or EXCELLENT!

danmash: @rodriguesk I'm not a native speaker but the distinction between "average" and "good" is more obvious that the distinction between "ok" and "good". 
These ration names are for developers only, so we should keep it clear. These names not in use in the frontend.

rodriguesk:
I am a native speaker and I'm saying AVERAGE means something different. If
OKAY isn't descriptive enough then we need another word. AVERAGE is
misleading. MEDIOCRE is probably better. It means "of only ordinary or
moderate quality; neither good nor bad; barely adequate"

rodriguesk:
This is an Net Promoter Score system. 3 needs to reflect it's a 'detractor'
in the net promoter scoring. The key is important. It's not just for
developers. It's for analysts too.

rodriguesk:
[image: image.png]

danmash:
okay

rodriguesk:
@danmash this isn’t passive the cypress tests currently?

app/static/Climate-Mind_bundled.yml
cyplas: I took a look, but I think there's a little confusion with which route we are talking about: @danmash slipped in an "s" in the pull request title, and @rodriguesk referenced the /conversations route, which we're not working on here. For /conversation, I double checked again and only found a mistake in conversationStatus for one of the examples, which is deprecated anyway.  I tweaked a couple more things, but quite minor.

cyplas:
> @cyplas can you please review the docs for this PR too? I don’t feel confident enough in me judging this only. We need to make sure sure the logic is perfect for what the frontend needs for the GET output and the PUT input requirements as we discussed in our 1-on-1 call a while back. That’s what matters most.

As far as I can tell, the changes haven't changed the basic API behaviour from what I implemented, but they did make the code significantly cleaner, error-checked, testable and tested. Looks good to me.

cyplas:
> It's because of Nick's requirements and update I made yesterday. Cypress should be adjusted

Is this why we're still getting cypress test fails? Does this mean that we shouldn't merge? I think this is the only thing left to be done in this PR.

PR #345, opened 2022-06-22
merge workflows to avoid CD without CI jobs

danmash:
run CD without CI doesn't make sense, the project should be tested first


danmash:
@rodriguesk it runs in parallel, I think it's a problem
Screenshot 2022-06-23 at 00 04 09

rodriguesk (approved):

PR #348, opened 2022-06-24
CM 1057 cleaner circleci and docker caching

rodriguesk:
Added support for:
- cleaner logs for pytest (separate section in circle ci report)
- docker caching (DLC = docker layer caching)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,14 +24,6 @@ jobs:
           name: "Build and Run all containers"
           command: |
             env DOCKER_BUILDKIT=0 docker-compose --file docker/docker-compose.pytest.yml up --build --abort-on-container-exit
-  build:
-    machine: true
-    steps:
-      - checkout
-      # build the application image
-      - run:
-          name: Docker Build
-          command: docker build -f docker/services/web/Dockerfile -t $DOCKER_REGISTRY_NAME/climatemindbackend:$CIRCLE_BUILD_NUM .
   cypress:
     # Set up test environment and run cypress tests
     machine:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,9 @@ jobs:
           name: Linter
           command: python3 -m black --check ./
   pytest:
-    machine: true
+    machine:
+      image: ubuntu-2004:202201-02
+      docker_layer_caching: true
     steps:
       - checkout
       - setup_remote_docker:
@@ -41,10 +43,9 @@ jobs:
     # Set up test environment and run cypress tests
     machine:
       image: ubuntu-2004:202201-02
+      docker_layer_caching: true
     steps:
       - checkout
-      - setup_remote_docker:
-          docker_layer_caching: true
       - run:
           name: Install Node Dependencies
           command: npm i
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,9 +17,7 @@ jobs:
           name: Linter
           command: python3 -m black --check ./
   pytest:
-    machine:
-      image: ubuntu-2004:202201-02
-      docker_layer_caching: true
+    machine: true
     steps:
       - checkout
       - run:
@@ -40,7 +38,6 @@ jobs:
     # Set up test environment and run cypress tests
     machine:
       image: ubuntu-2004:202201-02
-      docker_layer_caching: true
     steps:
       - checkout
       - run:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -17,25 +17,31 @@ jobs:
           name: Linter
           command: python3 -m black --check ./
   pytest:
-    machine: true
+    machine:
+      image: ubuntu-2004:202201-02
+      docker_layer_caching: true
```
danmash, line 22: This one works great! Major improvement!

Screenshot 2022-06-25 at 17 04 22

.circleci/config.yml
```diff
@@ -17,25 +17,31 @@ jobs:
...
-          command: docker build -f docker/services/web/Dockerfile -t $DOCKER_REGISTRY_NAME/climatemindbackend:$CIRCLE_BUILD_NUM .
+          name: "Docker Compose Up"
+          command: docker-compose --file docker/docker-compose.pytest.yml up -d
+      - run:
+          name: "Wait for database to spin up"
+          command: sleep 30
+      - run:
+          name: "Run Pytest"
+          command: docker exec -it docker-web-1 pytest --cipdb
+          # or to NOT check for breakpoints accidentally left in the code
```
danmash, line 38: Don't see a reason for this comment

.circleci/config.yml
```diff
@@ -17,25 +17,31 @@ jobs:
...
+          # docker-compose exec -it docker-web-1 pytest -n auto
```
danmash, line 39: wrong command, `-n auto` is for a parallel run which MSSQL doesn't support.

.circleci/config.yml
```diff
@@ -17,25 +17,31 @@ jobs:
...
   cypress:
     # Set up test environment and run cypress tests
     machine:
       image: ubuntu-2004:202201-02
+      docker_layer_caching: true
```
danmash, line 44: This doesn't work for cypress for some reason, it was faster before. So I suggest to fix or remove

docker/services/web/pytest/entrypoint.pytest.sh
```diff
@@ -3,8 +3,11 @@
...
+# python
+flask run --host=0.0.0.0
+
 # FIXME: unable to run pytest in parallel with xdist plugin due to the error
 #  Transaction (Process ID 52) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction. (1205) (SQLExecDirectW)')
 # pytest -n auto --cipdb
 
-pytest --cipdb
\ No newline at end of file
+# pytest --cipdb
```
danmash, line 13: commenting on everything is a bad practice. Just remove all comments if no longer needed

docker/services/web/pytest/entrypoint.pytest.sh
```diff
@@ -3,8 +3,11 @@
 echo "sleeping entrypoint"
 sleep 120
 
+# python
+flask run --host=0.0.0.0
```
danmash, line 7: to be honest, we don't need any command in this container at all, especially our web app
so this doesn't make sense. 
I suggest to remove entrypoint script at all 
and use `sleep infinity` instead

add the following line here `docker/docker-compose.pytest.yml` to the `web` service.
 
    command: sleep infinity

danmash (changes requested):
Do not merge without changes! And squash

.circleci/config.yml
rodriguesk: I think there was just a weird issue with Cypress that day. It was sitting there for a while doing nothing as the clock was ticking up.

docker/services/web/pytest/entrypoint.pytest.sh
rodriguesk: Yes sleep infinity should work too.

.circleci/config.yml
rodriguesk: TSQL and the MSSQL server should support some degree of parallelism though. It should allow multiple reads from what I remember but only 1 write at a time. I think the server should manage it though and make a queue. But given how fast it currently is (15sec) I don’t think parallelizing is even needed.

.circleci/config.yml
rodriguesk: To make sure that’s actually what’s happening. I wanted to ask you about that. Does it do that?

.circleci/config.yml
danmash: @rodriguesk don't need to remove it, it works fine for the `pytest` step

PR #367, opened 2022-08-01
CM-1064 delete conversation endpoint added

danmash:


rodriguesk:
@danmash I think it would be better to call 'isHidden' a better name that more suggests that the user has deactivated the conversation. Could you change the 'isHidden' variable to be 'isDeactivated' or 'isTrashed' instead? I think it is also better to do this so it's clearer how the unique link becomes deactivated and will cause error message on frontend if anyone tries to load that user B conversation link.

danmash:
It's all about perspective. If you look at the feature from the link perspective - maybe it's better to name the field "is_deactivated". But if you look at the feature from the conversation object (for example index page, list of conversations) - it's definitely hidden. This field belongs to objects named "conversations" (db table, python class), not "links". that's why for me is preferable to look at the feature from the conversation object perspective. 
In addition, I don't think the value of such a controversial improvement is worth the time I'll spend on this.

rodriguesk:
Ok, I agree about the name should be specific for the Python object. The only reason why we are keeping the record at all is for analytics, so it needs to be clear for analytics what this actually is and how this connects to user behavior. We aren’t keeping the record to resurrect it later, which is what it sounds like when it’s marked as hidden. The conversation isMarkedDeleted. I’m happy to update it for you if you’re short on time.

danmash:
You could push your suggestions to the branch

danmash:
@rodriguesk  please squash your commits into single one

rodriguesk:
I don’t have time for this. When you merge to develop it gets squashed anyways. Thanks for your help!

rodriguesk (approved):

danmash:
> When you merge to develop it gets squashed anyways.

working on the same branch together it's better not to squash on merge to distinguish changes. 
I'll do "rebase and merge" instead which will not squash all commits into a single one. 

I've squashed your commits manually

rodriguesk:
@danmash the frontend is having a hard time running this branch. Why aren’t the tests passing?

rodriguesk (approved):

PR #368, opened 2022-08-02
Use test app link

rodriguesk:
Use test app url for now so that beta testers aren’t directed to login to main app, but rather directed to login to the test app. This will need to be switched to production url when getting merged to master later on.


danmash:
@rodriguesk url should not be hardcoded at all, please create a task for you to change it back before the launch. Otherwise it could be a critical bug.

rodriguesk:
Yes, I agree. It was just fine for now so people can use the app. Non-feature technical tasks might be best for someone else to add in Jira so the details are correct. @danmash could you add it? I can include it in the next sprint we do.

PR #369, opened 2022-08-02
Test site url for login

rodriguesk:
Update welcome email to direct to test website for beta testing. Will need to be changed later to direct to production website when this code gets merged to the master branch.


danmash:
@rodriguesk such things shouldn't be hardcoded, please create a task to revert this before the launch.

PR #372, opened 2022-08-14
CM-1076 user_b_shared_email.py link changed

danmash:


rodriguesk (approved):

danmash:
@rodriguesk I have some problems with sendgrid on my side. don't have time to check right now. can you try and make sure that email sending is works properly?

rodriguesk:
The best way for us to check this is by merging to the develop branch. It should work.

danmash:
It could not work due to wrong type passed to send grid. Anyway, you could
test it how you prefer

PR #380, opened 2022-09-09
Cm 1110 get quiz id endpoint

rodriguesk:


cyplas (approved):
Maybe it would be nicer one day to make a single GET route with for info from the users table, but for now this seems fine to me. I just tweaked the stoplight headings style in line with our others.

Also, I wonder if we shouldn't be lower-casing the quizId to be in line with what uuid.uuid4() returns, but then it would be different then what MSSQL returns, and maybe that's a general thing we should standardise separately.

PR #420, opened 2022-10-19
updated copy for running the app locally

rodriguesk:
### Detailed information:

The copy should reflect how to run the app locally. Some people may not be contributing to changes, rather just trying to learn from us or explore. Or could be front end people just trying to run the backend. The copy should help let these people know where to find the instructions for running the app locally.t

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ Our [web application](https://app-frontend-prod-001.azurewebsites.net/) makes co
 
 The full API documentation to use on a front end including methods and request examples could be found [here](https://climatemind.stoplight.io/docs/climatemind-backend/).
 
-## How to contribute?
+## How to run the app locally and/or contribute?
 
-If you are interested in participating as a Python developer see our [contributor guide](docs/contribute-as-a-python-dev/).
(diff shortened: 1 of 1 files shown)


danmash (commented):
To be honest, I think it doesn't make sense because the goal is to contribute, not set up the local version. 

The local version setup is a part of the contribution. So you actually bring up subtask before the contribution.

danmash:
Also, see the guide about how to change the documentation. https://contribute.climatemind.org/v/rest-api/contribute-as-a-python-dev/how-to-update-docs/documentation

I know it's not obvious that we should not push any changes directly to the develop branch, but otherwise gitbook unable to keep up the up to date copy of the repo. 

So, if you would like to continue to propose your changes you should use gitbook "submit for review" or create the same PR to merge to `docs` instead of `develop` branch.

danmash:
@rodriguesk I've also created a guide for the documentation change through GitHub. This way is more convenient to use for such small changes like you suggested in this PR.  

https://contribute.climatemind.org/v/documentation/

danmash:
> we should not push any changes directly to the develop branch, but otherwise gitbook unable to keep up the up to date copy 

I'll try to fix it again later

PR #430, opened 2022-10-27
 #403: transformed e2e scores tests from cypress to pytest

cyplas:
### Detailed information:

Rewrote the scores e2e tests from cypress to python.
Refactored uuid validation to increase reusability.

### Closing issues: 

List all issues the pull request solve: 

- *#403*

### Test plan (required)

- [X] The PR contains new PyTest unit/integration tests for any function or functional added. 
- [X] The PR changes existing PyTest unit/integration tests to keep all tests up to date.
- [X] The PR does not lead to degradation in unit test coverage.
- [X] Project parts affected by changes in this PR was tested manually on your local (using Postman or in any other way). List everything you've tested below:
  - pytest


app/scoring/tests/test_scores_routes.py
```diff
@@ -0,0 +1,34 @@
+import uuid
+import os.path
+import pytest
+from flask import url_for
+from app.common.static import get_dict_from_json_file
+
+JSON_DIRECTORY = (
+    "cypress/fixtures"  # TODO: move to suitable python directory and make robust
```
cyplas, line 8: Regarding this TODO, I haven't moved the files yet, as they are used by other cypress tests yet to be rewritten.

app/scoring/tests/test_scores_routes.py
```diff
@@ -0,0 +1,34 @@
...
+)
+
+
+def is_uuid(string):
```
danmash, line 12: I'm sure we have some function to check if uuid is valid or no, no need to create a new one

danmash (changes requested):
I appreciate you've checked examples in the repo and used some stuff from it like "client_with_user_and_header". Nicely done 👍

danmash:
Hi @cyplas 
will you be able to adjust this?

app/scoring/tests/test_scores_routes.py
cyplas: > I'm sure we have some function to check if uuid is valid or no, no need to create a new one

We do have app.common.validate_uuid, but it raises one of our CustomErrors, which have HTTP status_codes, which are not relevant here. But perhaps that is too conservative and I should just go ahead and use it? In principle I agree with you.

app/scoring/tests/test_scores_routes.py
cyplas: Good idea, I've copied them to under common/, as they are also used by other e2e tests.

app/scoring/tests/test_scores_routes.py
danmash: So basically you have the same logic `try-except + uuid.UUID(string)`. It's a code duplication.
I recommend you move your new function to the common folder and use it here and in `app.common.validate_uuid`.

cyplas: Agreed. Finally addressed this in today's commits.

I did notice something else, however, but it seems out of scope here and I suggest creating a separate issue: app.common.uuid.validate_uuid (which is used in many places) takes two arguments, but its checks (only) apply to the first one and it's error messages (only) refer to the second one.

cyplas: > I did notice something else, however, but it seems out of scope here and I suggest creating a separate issue: app.common.uuid.validate_uuid (which is used in many places) takes two arguments, but its checks (only) apply to the first one and it's error messages (only) refer to the second one.

Though I don't know, maybe this was intentional, as the second argument does make the error messages a little more precise ...

danmash (approved):
Awesome @cyplas! 
The single thing I would like to change is moving .json files to `app/scoring/tests/data` since you are using them in the single test at the moment, these files are not common. You are free to merge after fixing this without new approval.

Issue #437, opened 2022-11-02
Fix services.webapp.environment.CHOKIDAR_USEPOLLING contains true, which is an invalid type, it should be a string, number, or a null


rscarrasco:
### Describe the bug

I just forked the repo, and cloned it in my computer. I then proceeded to [run it locally](https://github.com/rscarrasco/climatemind-backend/blob/develop/docs/contribute-as-a-python-dev/installation.md#up-containers), but when I issued:

```
docker-compose -p climatemind-backend --profile webapp -f docker/docker-compose.yml up -d --build
```

it threw the following:

```
ERROR: The Compose file './docker/docker-compose.yml' is invalid because:
services.webapp.environment.CHOKIDAR_USEPOLLING contains true, which is an invalid type, it should be a string, number, or a null
```

My environment:

  * Docker  20.10.12.
  * Docker Compose 1.29.2
  * Linux Mint 21

### To Reproduce

Steps to reproduce the issue:

- `cd` into the project directory.
- run the following:

```
docker-compose -p climatemind-backend --profile webapp -f docker/docker-compose.yml up -d --build
```

### Expected behavior

I expected no error message from the above command.

### Task

- [ ] Check out if quoting the `true` literal at  `CHOKIDAR_USEPOLLING: true` (file `docker/docker-compose.yml`) fixes the problem.

danmash:
@rscarrasco nicely done!
I am unable to check this at the moment since I use M1 mac and can't build `docker/docker-compose.yml`
Could you please check this quoting and let me know? You could also create a PR which will be merged ASAP.
Anyway, thank you! I really appreciate this 👍

rscarrasco:
@danmash thanks!

I was trying just that...I tried quoting, and it seems to be working just fine. I then tried something else: removing it altogether. I issued:

```
grep -ri chokidar .
```

and 

```
find . -name "*chokidar*"
```
but nothing was found. This leads me to wonder if Chokidar is even used at all. I tried removing it from the `docker/docker-compose.yml, deleting all related images/containers, and rebuilding. The system ran as before.

I'm not sure if I can remove it, but I'll send a PR removing Chokidar anyway.

danmash:
You are probably right. If I understand correctly it is needed only for [development](https://stackoverflow.com/a/44643246/2207154) and we can omit it since we don't change any files and use the ready-to-go docker image. Please correct me if I'm wrong. 
@Svenstar74 please correct me if I'm wrong.

PR #438, opened 2022-11-02
fix: Removes Chokidar from Docker Compose files

rscarrasco:
### Detailed information:

Attempting to run Docker Compose is throwing an error:

```
ERROR: The Compose file './docker/docker-compose.yml' is invalid because:
services.webapp.environment.CHOKIDAR_USEPOLLING contains true, which is an invalid type, it should be a string, number, or a null
```

Since it seems that Chokidar is not being using at all, this commit removes it from the Docker Compose files.

### Closing issues: 

- Closes *#437*

### Test plan (required)

- [X] The PR contains new PyTest unit/integration tests for any function or functional added. 
- [ ] The PR changes existing PyTest unit/integration tests to keep all tests up to date.
- [X] The PR does not lead to degradation in unit test coverage.
- [X] Project parts affected by changes in this PR was tested manually on your local (using Postman or in any other way). List everything you've tested below:
  - This affects the containers deployed, and no functionality at all.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -41,8 +41,6 @@ services:
       - api
     ports:
       - "3000:3000"
-    environment:
-      CHOKIDAR_USEPOLLING: true
     profiles:
       - webapp
 
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -47,3 +47,4 @@ services:
 volumes:
   mssql_data:
     driver: local
+
--- a/docker/docker-compose.m1.yml
+++ b/docker/docker-compose.m1.yml
@@ -42,8 +42,6 @@ services:
       - api
     ports:
       - "3000:3000"
-    environment:
-      CHOKIDAR_USEPOLLING: true
     profiles:
       - webapp
 


docker/docker-compose.yml
```diff
@@ -41,11 +41,9 @@ services:
...
-    environment:
-      CHOKIDAR_USEPOLLING: true
     profiles:
       - webapp
 
 volumes:
   mssql_data:
-    driver: local
\ No newline at end of file
+    driver: local
```
danmash, line 49: @rscarrasco could you please clarify this change? Is the docker build not working on your side without it? Or is something wrong with the database?

danmash: it's [probably](https://docs.docker.com/storage/volumes/#start-a-service-with-volumes) `local` by default. so no need to specify it

rodriguesk:
@rscarrasco <https://github.com/rscarrasco> you using powershell? You
should try that.

danmash:
@rodriguesk this change is only about the environment variable CHOKIDAR_USEPOLLING which will not affect the build process on different platforms. We are free to change env variables, I see no way it will affect Windows users somehow

docker/docker-compose.yml
rscarrasco: Ummm, I don't recall making that change. Its probably a mistake. I'll add it back.

rodriguesk:
See https://www.docker.com/blog/speed-up-your-development-flow-with-these-dockerfile-best-practices/

”Keep live reload active between the host and the container”

docker/docker-compose.yml
rscarrasco: It seems I accidentally removed a newline at the end of the file. Sorry About, fixing now.

rscarrasco:
@rodriguesk Hi! No, I'm not using PowerShell. I'm on a Linux box. You mean this Docker Compose file should be also tested in PowerShell?

> ”Keep live reload active between the host and the container”

Ok, so Chokidar is probably needed. I believe this PR should be rejected, and I should figure out something else. I'll read the post you've sent before making more changes.

rodriguesk:
I’m not certain about the situation with Linux. I haven’t tested this with
Linux so it could be there are errors or something with Linux.

danmash:
> ”Keep live reload active between the host and the container”
> Ok, so Chokidar is probably needed. I believe this PR should be rejected, and I should figure out something else. I'll read the post you've sent before making more changes.

@rodriguesk @rscarrasco guys, you just miss the main idea of the frontend `webapp` image in this repository :)  We are not going to change any JS files here. We just need it to test the backend.  It's a short track for backend developers to avoid the need for npm setup and building frontend image from scratch.

The PR is still valid and is to be merged.

> You mean this Docker Compose file should be also tested in PowerShell?

yes, in general. But this change looks safe enough.

docker/docker-compose.yml
danmash: Ah I see, it's about the new line only. I though you'd added the driver:local my bad.

danmash (approved):
Thank you for testing the docker build and fixing the redundant env variable!

rodriguesk:
If no coding changes are being made then it’s fine to remove this Chokidar. 
But this error suggests the frontend linux configuration settings are bugged and it should get fixed!

danmash:
I've just enabled CircleCI runs for external forks but it's still not running.
Should work fine next time. 
Screenshot 2022-11-02 at 16 53 21
This time I'll check without checks.

danmash:
@rscarrasco I [squashed](https://contribute.climatemind.org/v/rest-api/contribute-as-a-python-dev/development/making-a-pr#before-requesting-a-review) your commits into a single one. I hope you won't be mad about losing your authorship :)

Thanks again for your contribution, hope to see you again! Feel free to comment on any issue or start a [discussion](https://github.com/ClimateMind/climatemind-backend/discussions).

PR #465, opened 2022-11-30
Use localhost for base_frontend_url in tests

brianpeiris:
### Detailed information:

In ec87a78358d9e70b92e95d81cd633b1ecc277f51, we configured the local docker compose file to use locahost:3000 as the base_frontend_url. This breaks tests locally, since the tests expect the base_frontend_url to be app.climatemind.org.

This changes the tests to conform to localhost:3000, and it also changes `docker-compose.pytest.yml` so that it uses localhost in CI as well.

### Test plan (required)

- [x] The PR contains new PyTest unit/integration tests for any function or functional added. 
- [x] The PR changes existing PyTest unit/integration tests to keep all tests up to date.
- [x] The PR does not lead to degradation in unit test coverage.
- [x] Project parts affected by changes in this PR was tested manually on your local (using Postman or in any other way). List everything you've tested below:
  - Ran all pytests

### DIFF
--- a/app/tests/test_models.py
+++ b/app/tests/test_models.py
@@ -46,7 +46,7 @@ def test_password_reset_expired_property():
 
 def test_password_reset_url_with_default_url():
     password_reset = PasswordResetLinkFactory()
-    assert password_reset.reset_url.startswith("https://app.climatemind.org")
+    assert password_reset.reset_url.startswith("http://localhost:3000")
 
 
 @mock.patch("app.models.current_app")
--- a/app/tests/test_models.py
+++ b/app/tests/test_models.py
@@ -44,11 +44,6 @@ def test_password_reset_expired_property():
     assert not password_reset.expired, "PasswordResetLink should not be expired"
 
 
-def test_password_reset_url_with_default_url():
-    password_reset = PasswordResetLinkFactory()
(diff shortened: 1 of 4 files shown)


app/tests/test_models.py
```diff
@@ -46,7 +46,7 @@ def test_password_reset_expired_property():
 
 def test_password_reset_url_with_default_url():
     password_reset = PasswordResetLinkFactory()
-    assert password_reset.reset_url.startswith("https://app.climatemind.org")
+    assert password_reset.reset_url.startswith("http://localhost:3000")
```
danmash, line 49: yeah, but the default `test_password_reset_url_with_default_url` should be the case when we don't set the variable. In this case, there should be `https://app.climatemind.org`. We trying to simulate the production env this way. Add some mocks to this function

danmash (changes requested):
Nice catch!
Some changes needed

app/tests/test_models.py
brianpeiris: As discussed, since it would be difficult to test the environment variable behavior, I've just removed this test, since it is redundant.

danmash (approved):
Good job 👍

Issue #401, opened 2022-10-12
Replace cypress questions.cy.js with PyTest integration tests.


danmash:

tadakongithub:
I'd like take on this task. Is there any deadline?

danmash:
Hi @tadakongithub! Thank you for your interest, I really appreciate this 👍
There's no deadline. I'll send a follow-up in the comments in some time if no progress.
Feel free to ask any questions here or via chat https://t.me/climatemind_chat

tadakongithub:
Ok.

tadakongithub:
Hi. I'm a bit new to making tests and python app in general. While I'm writing in vs code, it says import "pytest" could not be resolved. Same goes for flask. Can you advise what to do?

danmash:
@tadakongithub to make able to run the project and pytest you need to setup your docker environment first. Please make sure you follow [the guide ](https://contribute.climatemind.org/v/rest-api/contribute/installation)
Let me know if you have any problems with the installation.

Then you should be able to [run any command](https://contribute.climatemind.org/v/rest-api/contribute/development/work-with-docker#execute-command-in-docker-container) in your docker container and connect to running containers [from your IDE](https://code.visualstudio.com/docs/devcontainers/containers#_attach-to-existing-container)

tadakongithub:
Just made a PR. Let me know if I need to make any changes

tadakongithub:
I used to be able to create a pull request from a working branch of my forked repo to the develop branch of the upstream repo. Now I can't seem to do that. Did something change? Should I push directly to the develop branch of the upstream repo?

brianpeiris:
@tadakongithub You should be able to create multiple pull requests from your fork, but they have to be on unique branches with different names. Is that what you're running into?
I don't have full access to the repo settings here, but I assume nothing changed there. I'm still able to create PRs from my fork.

brianpeiris:
We can close this now that #477 is merged. Thanks!

tadakongithub:
Now that it's merged, the pull request screen looks how it originally looked and it seems I can make a new PR now. I wonder if I cannot make multiple PR from my fork. Or maybe it was simply because my added commits went to the same PR I made.

Thanks for reviewing!

brianpeiris:
You should definitely be able to open multiple PRs from your fork. I've done that a few times for this project. They just have to be from separate branches. 
Let us know if you run into that issue again.

PR #477, opened 2022-12-10
test was written for questions in pytest to solve the issue #401

tadakongithub:
### Detailed information:

Wrote a test in pytest in order to test the endpoint "/questions" in place of "questions.cy.js".

### Closing issues: 

List all issues the pull request solve: 

- *#401*

### Test plan (required)

- [ ] The PR contains new PyTest unit/integration tests for any function or functional added. 
- [ yes] The PR changes existing PyTest unit/integration tests to keep all tests up to date.
- [ no] The PR does not lead to degradation in unit test coverage.
- [ yes] Project parts affected by changes in this PR was tested manually on your local (using Postman or in any other way). List everything you've tested below:
  - *Part 1*
  - *Part 2*


app/questions/tests/test_questions_routes.py
```diff
@@ -7,6 +7,75 @@
...
+    set_one = response.get_json()["SetOne"]
+
+    assert isinstance(set_one, list)
+    assert len(set_one) == 10
+    for question_obj in set_one:
+        assert type(question_obj) == dict
+        assert "id" in question_obj
+        assert "value" in question_obj
+        assert "question" in question_obj
+        assert type(question_obj["id"]) == int or type(question_obj["id"]) == float
```
brianpeiris, line 35: This should always be an `int`, so it would be better to remove the `float` comparison so that the test is stricter.

Same for the `"id`" assertions on like 52 and 68.

app/questions/tests/test_questions_routes.py
```diff
@@ -7,6 +7,75 @@
...
+def test_directions_properties(client):
+    response = client.get(url_for("questions.get_questions"))
+    direction = response.get_json()["Directions"]
+
+    assert type(direction) == str
+    """
     json_file_name = current_app.config.get("SCHWARTZ_QUESTIONS_FILE")
     with open(json_file_name) as json_file:
-        assert response.json == load(json_file)
+        assert response.json == load(json_file)"""
```
brianpeiris, line 81: It may still be worth keeping this test code, since it was already in place. It does a more thorough check of the values in the json response. I'd suggest keeping it under a test called `test_get_questions_matches_json_file`

app/questions/tests/test_questions_routes.py
```diff
@@ -7,6 +7,75 @@
...
+        assert "id" in answer
+        assert "text" in answer
+        assert type(answer["id"]) == int or type(answer["id"]) == float
+        assert type(answer["text"]) == str
+
+
+@pytest.mark.integration
+def test_directions_properties(client):
+    response = client.get(url_for("questions.get_questions"))
+    direction = response.get_json()["Directions"]
```
brianpeiris, line 75: Minor correction, this should variable should be plural, to match the json property name.
```suggestion
    directions = response.get_json()["Directions"]
```

brianpeiris (changes requested):
Made some suggestions, but it mostly looks good.
As @danmash mentioned, you can also delete the `questions.cy.js` file as part of this PR, and also delete `questionsEndpoint` from `cypress/support/commands.js`.

tadakongithub:
Noted all above. Let me work on the suggestions.

brianpeiris (approved):
LGTM! Thanks!

PR #483, opened 2023-02-05
Add CORS for the feedback route

Svenstar74:


danmash:
@Svenstar74 Could you please install black "pip install black" and lint all files with "black ."
Thanks

danmash (approved):

PR #497, opened 2023-08-05
Skip Recaptcha Validation for Native Apps

Svenstar74:
The Recaptcha is a good method for the website to counter scripting / bots. However, for our native apps on Android and iOS we don't like to have the Recaptcha. An optional skip_recaptcha parameter in the body allows to bypass the validation. Of course this is not an optimal solution, but for my estimation it is sufficiently for a small user base like we have it currently. As all the source code is open sourced, security plays an important role and we will improve it consistently.

### DIFF
--- a/app/auth/routes.py
+++ b/app/auth/routes.py
@@ -76,7 +76,7 @@ def login():
     if not user or not user.check_password(password):
         raise UnauthorizedError(message="Wrong email or password. Try again.")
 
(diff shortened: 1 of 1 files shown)


rodriguesk:
The plan is now to have no recaptcha, and to later have recaptcha get triggered after a certain number of failed login attempts.

rodriguesk (approved):

PR #537, opened 2024-08-25
Pushing requirements update to the branch for the OAuth changes

epixieme:
Added the following to the requirements.txt to fix the sentry issue
google-auth==2.22.0
google-auth-oauthlib==1.0.0
google-auth-httplib2==0.1.0


rodriguesk:
I’ll approve this but I think it’d be best if you just add a simple pytest to ensure that the functions you need to run can access the libraries they need to run. That way the you’d know if this fixes the issues without needing to merge to develop first.

rodriguesk (approved):

